PR #3, opened 2025-08-19
fun: Rick roll Users in Dangerous Links Security Showcase

aroramrinaal:
Nice idea from [twitter](https://www.youtube.com/watch?v=dQw4w9WgXcQ): let's not miss the opportunity to Rick roll in that dangerous links demo

Updated the "Get rich quick!" link to point to the classic rick roll video on yt


haydenbleasel:
Absolutely required 🙏

PR #15, opened 2025-08-21
fix(streamdown): add displayName for Block

FranciscoMoretti:
Fix Block displayname. Component definition is missing display name (eslintreact/display-name)


haydenbleasel:
Legend, thanks @FranciscoMoretti 🙏

PR #16, opened 2025-08-21
fix(Response): language available on code element, not pre

FranciscoMoretti:
- Infer code language from the `className` on the code element (not from pre).
- The `className` prop on the component is a string. The `node.properties.classname` is an array of class tokens. We use `className` with a regex to extract the language.
- Default language plaintext. Happy to change this if JS is a better default

### DIFF
--- a/packages/streamdown/lib/components.tsx
+++ b/packages/streamdown/lib/components.tsx
@@ -1,10 +1,88 @@
-import { isValidElement } from 'react';
+import { CodeIcon } from 'lucide-react';
+import { isValidElement, useState } from 'react';
 import type { Options } from 'react-markdown';
 import type { BundledLanguage } from 'shiki';
-import { CodeBlock, CodeBlockCopyButton } from './code-block';
+import {
+  CodeBlock,
+  CodeBlockCopyButton,
+  CodeBlockRenderButton,
+} from './code-block';
+import { Mermaid } from './mermaid';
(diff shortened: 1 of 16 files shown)


packages/streamdown/lib/components.tsx
```diff
@@ -4,6 +4,8 @@ import type { BundledLanguage } from 'shiki';
 import { CodeBlock, CodeBlockCopyButton } from './code-block';
 import { cn } from './utils';
 
+const LANGUAGE_RE = /language-(\w+)/;
```
alephpiece: Use `/language-([\w-+]+)/` to match `-` and `+`, e.g., `c++`

FranciscoMoretti: Replaced with `language-([^\s]+)` to match any non-whitespace after `language-`. This handled also c#

packages/streamdown/lib/components.tsx
```diff
@@ -127,32 +129,25 @@ export const components: Options['components'] = {
       {children}
     </blockquote>
   ),
-  code: ({ node, className, ...props }) => {
+  code: ({ node, className, children, ...props }) => {
```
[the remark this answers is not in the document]
FranciscoMoretti: Updated

haydenbleasel:
Great work @FranciscoMoretti 🙏

PR #28, opened 2025-08-24
Update README.md for dev to build streamdown first

davidkpiano:
Adds the step to build the streamdown package first, since otherwise it won't find the module in the workspace.


haydenbleasel:
Great catch @davidkpiano - will try and find a fix eventually so it uses source code rather than build.

PR #42, opened 2025-08-26
chore: remove package-lock.json

17hz:
Remove unintended npm lockfile from pnpm monorepo. Prevent future reappearance


haydenbleasel:
Nice one, thanks @17hz 🙏

PR #207, opened 2025-11-07
Stabilize Streamdown contexts

FranciscoMoretti:
## Summary
- Memoize objects passed to contexts

## Testing
- Verified using the compiled lib in an app

### Prev (code blocks and actions re-render even after the block is completed)

https://github.com/user-attachments/assets/7ba91890-97f4-4f31-b4bf-e38229cf6223

### New (code block and actions re-render only when that block is being streamed)

https://github.com/user-attachments/assets/b33a4129-4291-4018-ad97-d7ec2d817265

### DIFF
--- a/.changeset/cold-zoos-design.md
+++ b/.changeset/cold-zoos-design.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Stabilize Streamdown contexts


haydenbleasel:
Thanks @FranciscoMoretti 💪

PR #266, opened 2025-11-28
chore: move unified from devDependencies to dependencies

yeliex:
## Description

Since `unified` has been imported in `markdown.tsx`, it should be listed in dependencies.

This change prevent `unified` bundled into dist, which may cause `CallableInstance` in unified be mangled info `IIFE`, and then be mangled into `null` in app bundle minify.

## Before
in build dist:
```
// unified/callable-instance.js
var an = (function (e) {
    let o = this.constructor.prototype, r = o[e], s = function () {
        return r.apply(s, arguments);
    };
    return Object.setPrototypeOf(s, o), s;
});

// unified/index.js
var So = {}.hasOwnProperty, je = class e extends an {}

// after app minified
class e extends null {}
```

### DIFF
--- a/.changeset/light-keys-beg.md
+++ b/.changeset/light-keys-beg.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+chore: move unified from devDependencies to dependencies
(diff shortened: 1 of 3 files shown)


haydenbleasel:
Great catch, thanks @yeliex 🙏

PR #356, opened 2026-01-19
Add bundled language aliases for JS/TS code fences

keldenl:
## Summary
- Normalize common code-fence aliases using Shiki’s bundled language metadata (js/cjs/mjs, ts/cts/mts, zsh, etc.) in @streamdown/code.
- Add tests covering alias support.

## Why
LLMs commonly emit "js" or "ts" fences by default. Mapping Shiki’s built-in aliases to bundled grammars keeps syntax highlighting working without prompt workarounds to force "javascript"/"typescript".

## Testing
- pnpm test
- pnpm run check
- pnpm run check-types

## Issue links
No existing issues found for alias-based highlighting after broad search (code block, highlight, syntax highlight, alias, js).

### DIFF
--- a/packages/streamdown-code/index.ts
+++ b/packages/streamdown-code/index.ts
@@ -86,7 +86,7 @@ const normalizeLanguage = (language: string): string => {
   if (languageNames.has(lower as BundledLanguage)) {
     return lower;
   }
-  return trimmed;
+  return lower;
 };
 
 // Singleton highlighter cache


keldenl:
Addressed the inline review note about casing: normalizeLanguage now lowercases unknown ids (while still mapping Shiki aliases). This keeps case handling consistent for Shiki lookups.

haydenbleasel:
Hey @keldenl, thanks for this PR! 🙌

Really nice approach using Shiki's `bundledLanguagesInfo` to derive aliases rather than hardcoding them - much more maintainable.

One small suggestion: the `normalizeLanguage` function could be simplified since the last two branches both return `lower`:

```typescript
const normalizeLanguage = (language: string): string => {
  const trimmed = language.trim();
  const lower = trimmed.toLowerCase();
  return languageAliases[lower] ?? lower;
};
```

Not a blocker though, will merge anyway as the logic is correct as-is. This solves a real pain point with LLM-generated code blocks!

PR #365, opened 2026-01-22
fix: footnote detection incorrectly matching regex character classes

thedonmon:
## Problem

The `parseMarkdownIntoBlocks` function in `packages/streamdown/lib/parse-blocks.tsx` has overly permissive footnote detection patterns that incorrectly match regex negated character classes.

### Current patterns (lines 4-5):
```typescript
const footnoteReferencePattern = /\[\^[^\]\s]{1,200}\](?!:)/;
const footnoteDefinitionPattern = /\[\^[^\]\s]{1,200}\]:/;
```

The `[^\]\s]` part matches **any character except `]` and whitespace**, which causes false positives.

### Example of the bug:

When parsing markdown containing a code block with regex patterns:

```markdown
# Regex Examples

\`\`\`perl
# Match URLs  
https?://[^\s<>"{}|\\^`\[\]]+
\`\`\`

More text here.
```

The pattern `[^\s<>"{}|\\^` in the code block is incorrectly detected as a footnote reference because:
- `\[` matches `[`
- `\^` matches `^`
- `[^\]\s]{1,200}` matches `\s<>"{}|\\` (any chars that aren't `]` or whitespace)
- `\]` matches `]`

This causes `parseMarkdownIntoBlocks` to return the **entire document as a single block**, breaking block-level parsing for documents containing regex patterns in code blocks.

## Solution

Update the footnote patterns to only match valid footnote identifiers. According to markdown footnote syntax, identifiers should be alphanumeric with underscores and hyphens (e.g., `[^1]`, `[^note]`, `[^my-note]`, `[^my_note123]`).

### Updated patterns:
```typescript
const footnoteReferencePattern = /\[\^[\w-]{1,200}\](?!:)/;
const footnoteDefinitionPattern = /\[\^[\w-]{1,200}\]:/;
```

Using `[\w-]` (word characters plus hyphen) instead of `[^\]\s]` ensures only valid footnote identifiers are matched.

## Changes

1. **`packages/streamdown/lib/parse-blocks.tsx`** (lines 4-5):
   - Changed `[^\]\s]` to `[\w-]` in both footnote patterns
   - Updated comments to clarify valid footnote identifier format

2. **`packages/streamdown/__tests__/footnotes.test.tsx`** (new tests):
   - Added regression tests for regex character classes in code blocks
   - Added tests to ensure real footnotes still work correctly

### Before fix:
```typescript
const markdown = `# Example\n\n\`\`\`\nhttps?://[^\\s<>]+\n\`\`\`\n\nMore text.`;
parseMarkdownIntoBlocks(markdown); // Returns: [entire document] (1 block)
```

### After fix:
```typescript
const markdown = `# Example\n\n\`\`\`\nhttps?://[^\\s<>]+\n\`\`\`\n\nMore text.`;
parseMarkdownIntoBlocks(markdown); // Returns: multiple blocks as expected
```

### Real footnotes still work:
```typescript
const markdown = `Text[^1].\n\n[^1]: Footnote content.`;
parseMarkdownIntoBlocks(markdown); // Returns: [entire document] (1 block) - correct!
```

## Checklist

- [x] Tests pass (`pnpm test`)
- [x] Code is properly formatted (`pnpm check`)
- [x] Changeset added
- [x] Regression tests added for the fix
- [x] Existing footnote tests still pass

### DIFF
--- a/.changeset/fix-footnote-regex.md
+++ b/.changeset/fix-footnote-regex.md
@@ -1,0 +1,9 @@
+---
+"streamdown": patch
+---
+
+Fix footnote detection incorrectly matching regex character classes
+
+The footnote reference and definition patterns were too permissive, using `[^\]\s]` which matches any character except `]` and whitespace. This caused regex negated character classes like `[^\s...]` in code blocks to be incorrectly detected as footnotes, resulting in the entire document being returned as a single block.
+
+Updated the patterns to only match valid footnote identifiers (alphanumeric characters, underscores, and hyphens) using `[\w-]` instead.
--- a/packages/streamdown/__tests__/footnotes.test.tsx
+++ b/packages/streamdown/__tests__/footnotes.test.tsx
@@ -1,6 +1,6 @@
 import { render } from "@testing-library/react";
 import { describe, expect, it } from "vitest";
-import { Streamdown } from "../index";
+import { parseMarkdownIntoBlocks, Streamdown } from "../index";
 
 describe("Footnotes", () => {
   it("should render footnote references and definitions correctly", () => {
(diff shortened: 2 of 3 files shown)


haydenbleasel:
Great fix, thanks @thedonmon 🚢

PR #366, opened 2026-01-22
fix: don't treat $$ inside code blocks as math delimiters

thedonmon:
## Problem

Code blocks can contain `$$` as shell syntax (e.g., `pstree -p $$` for current process ID in bash). The math block merging logic in `parseMarkdownIntoBlocks` was incorrectly counting `$$` inside code blocks, causing subsequent content to be merged as if it were part of an unclosed math block.

### Example of the bug:

```markdown
\`\`\`bash
pstree -p $$
\`\`\`

Some text after.
```

The `$$` in the code block was being counted by `countDoubleDollars()`, and if there was any subsequent `$$` in the document, blocks would be incorrectly merged.

## Solution

Added tracking of whether the previous token was a code block. The math block merging logic now skips merging when the previous block was a code block, since `$$` inside code blocks is legitimate shell syntax, not a math delimiter.

## Changes

1. **`packages/streamdown/lib/parse-blocks.tsx`**:
   - Added `previousTokenWasCode` tracking variable
   - Updated math merging conditions to skip when previous token was code
   - Track code blocks after processing each token (ignoring space tokens)

2. **`packages/streamdown/__tests__/dollar-sign.test.tsx`**:
   - Added tests for `$$` inside code blocks
   - Added test for code block followed by math block (ensures math still works)

## Testing

```typescript
const markdown = `\`\`\`bash
pstree -p $$
\`\`\`

Some text after.`;

const blocks = parseMarkdownIntoBlocks(markdown);
// Before: blocks might be incorrectly merged
// After: code block and text are properly separated
```

## Checklist

- [x] Tests pass (`pnpm test`)
- [x] Changeset added
- [x] Existing dollar-sign tests still pass
- [x] Math blocks still work correctly after code blocks

### DIFF
--- a/.changeset/fix-dollar-sign-code-blocks.md
+++ b/.changeset/fix-dollar-sign-code-blocks.md
@@ -1,0 +1,9 @@
+---
+"streamdown": patch
+---
+
+Fix $$ inside code blocks being treated as math delimiters
+
+Code blocks can contain `$$` as shell syntax (e.g., `pstree -p $$` for current process ID). The math block merging logic was incorrectly counting `$$` inside code blocks, causing subsequent content to be merged as if it were part of a math block.
+
+Added tracking of previous token type to skip math merging when the previous block was a code block.


haydenbleasel:
Fantastic PR, thanks @thedonmon 🙏

PR #400, opened 2026-02-17
fix(remend): don't treat underscores inside HTML tags as emphasis

### LINKED ISSUE #385: The remend function adds trailing underscore when processing <a> tags with target="_blank"
### Bug Description

When using remend on a string containing `<a>` tags with target="_blank", an unexpected trailing `_` is appended to the output.

### Steps to Reproduce

**Usage:**
```js
const link = '<a target="_blank" href="https://link.com">word</a>'
const result = remend(link);
```

**Output:**
```js
<a target="_blank" href="https://link.com">word</a>_
```

### Expected Behavior

No extra `_` at the end

### Actual Behavior

Extra `_` at the end

### Code Sample

```tsx
const link = '<a target="_blank" href="https://link.com">word</a>'
const result = remend(link);
```

### Streamdown Version

remend: 1.1.0

### React Version

18.3.1

### Node.js Version

22

### Browser(s)

_No response_

### Operating System

None

### Additional Context

_No response_

sleitor:
## What?

Fix trailing `_` being appended when `remend()` processes HTML tags containing underscores in attribute values (e.g. `target="_blank"`).

## Why?

The emphasis handler counts single underscores to detect unclosed italic markers. It already skips underscores inside markdown links and math blocks, but didn't account for HTML tags. So `<a target="_blank">` was parsed as having an unclosed `_` emphasis, and remend appended a closing `_`.

**Before:**
```
remend('<a target="_blank" href="https://link.com">word</a>')
// → '<a target="_blank" href="https://link.com">word</a>_'
```

**After:**
```
remend('<a target="_blank" href="https://link.com">word</a>')
// → '<a target="_blank" href="https://link.com">word</a>'
```

## How?

1. Added `isWithinHtmlTag(text, position)` utility — searches backwards from the underscore position to check if it's between `<` and `>`.
2. Added the check to `shouldSkipUnderscore()` in `emphasis-handlers.ts`.
3. Added test cases for `target="_blank"` and other HTML attributes with underscores.

All 283 tests pass.

Fixes #385

### DIFF
--- a/packages/remend/__tests__/html-tags.test.ts
+++ b/packages/remend/__tests__/html-tags.test.ts
@@ -62,6 +62,18 @@ describe("incomplete HTML tag stripping", () => {
     expect(remend("<div>Hello</div> <span")).toBe("<div>Hello</div>");
   });
 
+  it("should not add trailing underscore for HTML attributes with underscores", () => {
+    expect(remend('<a target="_blank" href="https://link.com">word</a>')).toBe(
+      '<a target="_blank" href="https://link.com">word</a>'
+    );
+    expect(remend('<a target="_blank">link</a>')).toBe(
+      '<a target="_blank">link</a>'
+    );
+    expect(remend('<iframe src="x" sandbox="allow_scripts">')).toBe(
+      '<iframe src="x" sandbox="allow_scripts">'
+    );
+  });
+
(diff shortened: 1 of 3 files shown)


haydenbleasel:
Nice catch @sleitor 🙏

PR #423, opened 2026-02-25
feat: add literalTagContent prop for plain-text custom tag children

### LINKED ISSUE #419: Feature request: opt-in literal text mode for custom tag children (`allowedTags`)
### Summary

Please add an opt-in way to render children of selected custom tags as plain text (no markdown parsing).

### Motivation

In AI UIs, custom tags are often used for structured entities such as mentions:

```html
<mention user_id="123">@_some_username_</mention>
```

Today, markdown in tag children is still parsed, so `_some_username_` can become italics.

For entity tags (usernames/handles/IDs), this is often unintended.

### Current Behavior

With `allowedTags` + `components`, child text is parsed as markdown (expected CommonMark/react-markdown behavior), e.g.:

- `_some text_` -> emphasis
- `@_foo_` -> emphasis in handle-like content

This happens even with `parseIncompleteMarkdown={false}` (because this is normal markdown parsing, not remend completion).

### Why this matters

For mention/entity tags, children are usually data labels, not prose. Requiring model-side escaping (`\\_`) everywhere is fragile and adds prompt/runtime complexity.

### Minimal Repro

```tsx

function Mention(props: any) {
  return <span data-tag="mention" {...props} />;
}

<Streamdown
  allowedTags={{ mention: ['user_id'] }}
  components={{ mention: Mention }}
>
  {`<mention user_id="123">_some text_</mention>`}
</Streamdown>
```

Actual output renders emphasis inside `mention` content.
Desired output is literal `_some text_`.

### Requested API (non-breaking, opt-in)

Any equivalent API would help, for example:

```tsx
<Streamdown
  allowedTags={{ mention: ['user_id'] }}
  literalTagContent={['mention']}
/>
```

or per-tag config:

```tsx
allowedTags={{
  mention: { attributes: ['user_id'], literalChildren: true }
}}
```

### Expected semantics for literal mode

1. Children text for specified tags is treated as plain text.
2. Markdown emphasis/link/code parsing is skipped for those children.
3. Default behavior remains unchanged for all other tags and users.

### Workaround today

Escaping underscores (`\\_`) works, but is brittle for AI-generated content and requires extra sanitization logic in app code.

An official opt-in mode would make mention/entity rendering much more robust.

sleitor:
## Summary

Adds an opt-in `literalTagContent` prop that renders children of specified custom tags as plain text (no markdown parsing). This is useful for mention/entity tags in AI UIs where child content is a data label (e.g. a username or handle) rather than prose.

## Motivation

When using `allowedTags` with custom components, markdown inside tag children is still parsed. For example:

```html
<mention user_id="123">@_some_username_</mention>
```

renders with emphasis on `_some_username_`, which is unintended for entity tags.

Requiring model-side escaping (`\_`) everywhere is fragile and adds prompt/runtime complexity.

## Solution

A new `literalTagContent?: string[]` prop. For tags in that list, markdown metacharacters inside the tag content are escaped before the markdown parser runs, so the parser treats the children as plain text.

```tsx
<Streamdown
  allowedTags={{ mention: ['user_id'] }}
  literalTagContent={['mention']}
>
  {`<mention user_id="123">@_some_username_</mention>`}
</Streamdown>
```

**Output:** renders `@_some_username_` as literal text — no emphasis.

## Why preprocessing, not a rehype plugin

The earlier approach (a rehype plugin that flattens children to text post-parse) loses the original characters — by the time rehype runs, `_some_username_` has already been parsed into `<em>some_username</em>` so the underscores are gone. Escaping before parsing preserves the original content.

## Changes
- **New** `packages/streamdown/lib/preprocess-literal-tag-content.ts` — escapes markdown metacharacters inside matching tag content before parsing
- **Modified** `packages/streamdown/index.tsx` — adds `literalTagContent?: string[]` to `StreamdownProps` and wires the preprocessor into `processedChildren`
- **Modified** `packages/streamdown/__tests__/allowed-tags.test.tsx` — 5 new tests for the `literalTagContent` prop
- **New** `packages/streamdown/__tests__/preprocess-literal-tag-content.test.ts` — 10 unit tests for the preprocessor

All 807 existing tests continue to pass.

Closes #419

### DIFF
--- a/.changeset/literal-tag-content.md
+++ b/.changeset/literal-tag-content.md
@@ -1,5 +1,5 @@
 ---
-"streamdown": patch
+"streamdown": minor
 ---
 
 Add `literalTagContent` prop for plain-text custom tag children
--- a/packages/streamdown/index.tsx
+++ b/packages/streamdown/index.tsx
@@ -768,6 +768,7 @@ export const Streamdown = memo(
     prevProps.className === nextProps.className &&
     prevProps.linkSafety === nextProps.linkSafety &&
     prevProps.normalizeHtmlIndentation === nextProps.normalizeHtmlIndentation &&
+    prevProps.literalTagContent === nextProps.literalTagContent &&
     JSON.stringify(prevProps.translations) ===
       JSON.stringify(nextProps.translations) &&
     prevProps.prefix === nextProps.prefix
--- a/packages/streamdown/lib/preprocess-literal-tag-content.ts
+++ b/packages/streamdown/lib/preprocess-literal-tag-content.ts
@@ -7,14 +7,31 @@
  * rehype plugins execute the markdown has already been parsed and structural
  * information (e.g. underscores around emphasis) is lost.
  *
+ * Note: content that already contains backslash escapes (e.g. `\_`) will be
+ * double-escaped and render with a visible backslash. This is acceptable since
+ * the intended use case is raw data labels (usernames, handles) that should
+ * not contain markdown escape sequences.
+ *
  * Example:
  *   Input:  `<mention user_id="123">_some_username_</mention>`
  *   Output: `<mention user_id="123">\_some\_username\_</mention>`
  *   Rendered: literal `_some_username_`
  */
 
-// All characters that CommonMark treats as potential markdown metacharacters
-const MARKDOWN_ESCAPE_RE = /([\\`*_{}[\]()#+\-.!|~])/g;
+// Characters that CommonMark treats as inline formatting metacharacters.
+// Only escapes characters that can trigger formatting anywhere in text:
+//   \  backslash escapes
+//   `  code spans
+//   *  emphasis / strong
+//   _  emphasis / strong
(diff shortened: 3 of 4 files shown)


sleitor:
Thanks for the automated review! The HTML comment corruption issue is fixed in the latest commit (465de11).

**Root cause:** `preprocessCustomTags` ran before `preprocessLiteralTagContent`. When a tag appeared in both lists and its content contained blank lines, the `<!---->` markers inserted by `preprocessCustomTags` were subsequently escaped to `<\!\-\-\-\->` by `preprocessLiteralTagContent` (which treats `!` and `-` as markdown metacharacters).

**Fix:** Swapped the execution order — `preprocessLiteralTagContent` now runs first, so the markdown escaping happens before the blank-line markers are inserted. The `<!---->` comments are never seen by the escaper.

haydenbleasel:
thanks @sleitor great ship 🙏

PR #434, opened 2026-03-06
fix(streamdown-code): fall back to plain text for unknown/truncated language identifiers

### LINKED ISSUE #433: I want to report a bug...
### Bug Description

When using `<Streamdown>` with the `streamdown-code` plugin for text streaming, if the model output's code block language identifier gets network-truncated (for instance, the model intends to output `rust`, but the current chunk only receives the string `rus`), the internal `normalizeLanguage` function inside `streamdown-code` returns this unrecognized/truncated language name `"rus"` unmodified to the Shiki engine.

Because Shiki cannot find a language package named `"rus"`, it throws an exception:
`ShikiError: Language rus is not included in this bundle.`

Although this exception is caught by a `.catch` block in `streamdown-code/index.ts` which logs a `console.error`, **it does not provide any graceful fallback rendering result (TokensResult)**, nor does it recover the Promise chain. This leads to an irreversible blockage in the React rendering cycle, causing a Fatal Error in the downstream React component tree (the streaming message bubble) and terminating the rendering of all subsequent streamed text.

### To Reproduce
1. Import and render `<Streamdown>` in a React project, configuring `plugins={{ code: codePlugin }}`.
2. **Critical Condition**: Simulate a slow stream by setting `chunk size` to `1`.
3. Start streaming a Markdown text. When the stream hangs at the truncated identifier ` ```rus `, before the final `t` of `rust` is received.
4. Observe Console: Immediately throws `[Streamdown Code] Failed to highlight code: ShikiError: Language rus is not included in this bundle.`
5. Observe UI: The React rendering pipeline breaks, resulting in a white screen for that area or a permanent halt in the rendering of that message.

### Expected Behavior

The renderer should gracefully fallback to plain `text` highlighting when it encounters an unknown or truncated language identifier like `rus`. This ensures the UI remains stable while waiting for the subsequent chunk to complete the identifier (e.g., to `rust`).

### Actual Behavior

When receiving the truncated ` ```rus `, the plugin fails to provide any fallback. It throws an unhandled error, causing the entire message bubble component to crash (blank screen) and preventing any further rendering.

### Code Sample

```tsx

```

### Streamdown Version

 Streamdown Version: 2.3.0 (@streamdown/code: 1.0.3)

### React Version

18

### Node.js Version

_No response_

### Browser(s)

_No response_

### Operating System

None

### Additional Context

_No response_

sleitor:
Fixes #433

## Problem

When a code block language identifier is network-truncated during streaming (e.g. ```rus`` before ```rust`` is fully received), Shiki throws:

```
ShikiError: Language rus is not included in this bundle.
```

The existing `.catch()` handler only logs the error and cleans up subscribers — it doesn't provide a fallback `TokensResult`. This leaves the render callback hanging, causing the React component tree to crash (blank screen).

## Fix

Before calling `getHighlighter()`, check whether the resolved language name is in the `languageNames` set. If it isn't recognized (truncated or unknown), substitute `'text'` so Shiki renders the block as plain uncolored text while the stream catches up.

```ts
const safeLang = languageNames.has(resolvedLanguage as BundledLanguage)
  ? (resolvedLanguage as BundledLanguage)
  : ('text' as BundledLanguage);
```

All existing tests pass.

### DIFF
--- a/packages/streamdown-code/index.ts
+++ b/packages/streamdown-code/index.ts
@@ -208,10 +208,10 @@ export function createCodePlugin(
 
       // Resolve language to 'text' if not supported (e.g. truncated identifier)
       const safeLanguage: BundledLanguage | SpecialLanguage = languageNames.has(
-        resolvedLanguage as BundledLanguage,
+        resolvedLanguage as BundledLanguage
       )
         ? (resolvedLanguage as BundledLanguage)
-        : 'text';
+        : "text";
 
       // Start highlighting in background
       getHighlighter(safeLanguage, themes)


haydenbleasel:
great fix @sleitor appreciate it 🚀

PR #463, opened 2026-03-12
fix(streamdown): increase dropdown z-index

ziyadev:
## Summary
Increases the z-index of the copy dropdown in copy-dropdown.tsx to ensure it appears above the sticky table header.

## Problem
In fullscreen mode, when hovering over a code block that contains a table, the copy dropdown is rendered behind the sticky <thead>, causing it to be partially or fully hidden. In normal (non-fullscreen) view the dropdown appears correctly.

## Fix
Increased the z-index on the copy dropdown so it always renders on top of the table header, regardless of view mode.

screenshot-2026-03-12_11-11-15

### DIFF
--- a/.changeset/fix-dropdown-z-index.md
+++ b/.changeset/fix-dropdown-z-index.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Increase dropdown z-index for table copy and download menus to prevent clipping by surrounding elements.


aradhyacp:
Nice catch

Could you update the PR description to mention that the issue occurs specifically when the code block is in **fullscreen mode**? In normal view the dropdown appears to work as expected, but the overlap seems to happen in fullscreen.

Adding that context to the PR body would help maintainers reproduce and verify the issue more quickly :)

ziyadev:
> Nice catch
> 
> 
> 
> Could you update the PR description to mention that the issue occurs specifically when the code block is in **fullscreen mode**? In normal view the dropdown appears to work as expected, but the overlap seems to happen in fullscreen.
> 
> 
> 
> Adding that context to the PR body would help maintainers reproduce and verify the issue more quickly :)

Ofc

haydenbleasel:
fantastic, thanks @ziyadev

Issue #464, opened 2026-03-13
Playground markdown textarea has incorrect height in Safari 18


hamster1963:
### Bug Description

The playground page currently relies on field-sizing: content for the left-side markdown input textarea.

<img width="2912" height="756" alt="Image" src="https://github.com/user-attachments/assets/f5e512c0-b454-48eb-a8d8-fc2adea9d1cf" />

This works in browsers that support field-sizing, but Safari 18 does not support it, so the textarea height is calculated incorrectly. As a result, the markdown input area does not size/fill the editor pane as intended, which breaks the layout and makes the editing experience inconsistent on Safari.

### Steps to Reproduce

1.Open the website playground in Safari 18.
2.Look at the left-side markdown input area.
3.Compare its height/behavior with Chromium-based browsers.

### Expected Behavior

The markdown textarea should fill the available height of the left editor pane and behave consistently across browsers.

<img width="3348" height="2016" alt="Image" src="https://github.com/user-attachments/assets/6af7881b-7211-4487-835b-000bdf8be05e" />

### Actual Behavior

In Safari 18, the textarea height is incorrect, so the left editor pane does not render properly.

### Code Sample

```tsx

```

### Streamdown Version

2.4.0

### React Version

19.0.0

### Browser(s)

Safari

### Operating System

macOS

PR #465, opened 2026-03-13
fix(playground-editor): add full height class to ConversationContent

hamster1963:
## Description

Fix the playground markdown input height in Safari 18 by avoiding `field-sizing: content` for that textarea and letting it fill the left editor pane with local layout overrides only.

## Type of Change

## Related Issues

Fixes https://github.com/vercel/streamdown/issues/464

## Changes Made

- Added `h-full` to the playground markdown input container so the editor pane has a stable height.
- Overrode the playground textarea to use fixed field sizing instead of `field-sizing: content`.
- Kept the fix scoped to the playground page to avoid changing textarea behavior elsewhere.

## Screenshots/Demos

Safari浏览器 2026-03-13 14 02 39

### DIFF
--- a/.changeset/fix-playground-safari-textarea.md
+++ b/.changeset/fix-playground-safari-textarea.md
@@ -1,5 +1,0 @@
----
-"streamdown": patch
----
-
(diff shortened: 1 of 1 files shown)


haydenbleasel:
Thanks @hamster1963 🙏

PR #466, opened 2026-03-15
fix(streamdown): move mermaid from devDependencies to dependencies

smorimoto:
## Summary

- Move `mermaid` from `devDependencies` to `dependencies` in the `streamdown` package
- The published type declarations (`dist/index.d.ts`) import `MermaidConfig` from `mermaid`, making it part of the public API
- Without this change, consumers could resolve different `mermaid` versions for `streamdown` and `@streamdown/mermaid`, causing type errors due to incompatible `MermaidConfig` types

## Test plan

- [x] `pnpm build:packages` succeeds
- [x] Verify that consumer projects no longer need `mermaid` overrides to avoid type mismatches

### DIFF
--- a/packages/streamdown/package.json
+++ b/packages/streamdown/package.json
@@ -49,7 +49,6 @@
     "@vitejs/plugin-react": "^5.1.2",
     "@vitest/coverage-v8": "^4.0.15",
     "jsdom": "^27.3.0",
-    "mermaid": "^11.12.2",
     "react-markdown": "^10.1.0",
     "rehype-parse": "^9.0.1",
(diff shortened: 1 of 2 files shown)


haydenbleasel:
Thanks @smorimoto 🚀

mcuelenaere:
Due to this change, mermaid is now a forced dependency and this pulls in ~100MB of transitive dependencies. Also, this introduces a ~1MB (ungzipped) build artificat.

Can this be implemented in a different way, so mermaid is not a dependency and downstream consumers do not have to suffer from its size, if they opt to not use it?

PR #474, opened 2026-03-22
fix(streamdown): render code blocks incrementally during streaming

### LINKED ISSUE #473: Fenced code blocks don't render incrementally during streaming
# Fenced code blocks don't render incrementally during streaming

## Description

When streaming AI responses token-by-token, fenced code blocks (triple backtick) appear to render all at once rather than incrementally as the markdown string grows. Text outside code blocks streams smoothly with visible character-by-character rendering, but the moment a code fence opens, the content inside appears to buffer until the block is complete, then dumps visually.

## Environment

- `streamdown`: 2.5.0
- `@streamdown/code`: 1.1.1
- React 18
- Using `<Streamdown animated isAnimating plugins={{ cjk, code, math, mermaid }}>`

## Reproduction

1. Pass a growing markdown string to `<Streamdown>` that includes a fenced code block
2. The string updates token-by-token (verified via `console.log` — length increments: 4, 8, 24, 26, 34, 58... smoothly up to 860+)
3. Prose text before and after the code fence streams beautifully with the `animated` caret
4. The code block content appears all at once or in large chunks

Minimal usage:

```tsx
<Streamdown animated isAnimating plugins={{ cjk, code, math, mermaid }}>
  {streamingMarkdownString}
</Streamdown>
```

Where `streamingMarkdownString` grows incrementally and contains:

````
Some text before...

```typescript
// This content appears to dump all at once
function example() {
  return "hello";
}
```
````

### 1. Transport is not the issue

Console logging confirms the raw string updates smoothly, token-by-token. The buffering is in the rendering layer, not the data transport.

### 2. `parseMarkdownIntoBlocks` uses `marked`'s `Lexer.lex()`

`parseMarkdownIntoBlocks` calls `Lexer.lex(markdown, { gfm: true })` and iterates tokens. When the lexer encounters a fenced code block, it emits a single `code` token only after the closing fence is found. While the fence is open (no closing backticks yet), the raw content likely falls into a different token type or gets merged differently.

**Key question: Does `marked`'s `Lexer.lex()` emit a `code`-type token for a partial fence (no closing backticks), or does it fall back to a `paragraph` token?** If it treats unclosed fences as paragraphs, the block would render as text during streaming and then suddenly swap to a code block when the closing fence arrives — which is exactly what we observe.

### 3. `isIncomplete` infrastructure exists

The Streamdown component checks for unclosed code fences (via `hasIncompleteCodeFence`) on the last block and passes `isIncomplete={true}` to the `Block` component. The `useIsCodeFenceIncomplete` hook exposes this. However, if `marked`'s lexer doesn't produce a code-type token for the unclosed fence, the `isIncomplete` flag may not trigger the code block rendering path.

### 4. `chatbot.ai-sdk.dev` achieves smooth code block streaming

The reference chatbot at chatbot.ai-sdk.dev (which uses Streamdown via MessageResponse) renders code blocks incrementally during streaming. This suggests either a configuration difference or that AI SDK's `useChat` delivers data in a way that interacts with Streamdown differently than a custom WebSocket stream.

### 5. `parseIncompleteMarkdown` prop

We noticed this prop defaults to `false`. Does `remend` processing affect how unclosed code fences are rendered during streaming? Does enabling `parseIncompleteMarkdown={true}` change the behavior?

## Our setup (not using AI SDK)

We're NOT using AI SDK's `useChat`. We have a custom WebSocket streaming pipeline that delivers text tokens from a container running Claude CLI. The growing markdown string is accumulated in a Zustand store and passed as children to Streamdown. The string IS raw markdown with fences.

## Questions

1. Does `parseMarkdownIntoBlocks` produce incremental block updates for unclosed code fences?
2. Is there a prop or configuration we're missing to enable incremental code block rendering?
3. Does `chatbot.ai-sdk.dev` use any additional configuration beyond what's in its open-source `MessageResponse` wrapper?
4. Could `remend` preprocessing be the key — does it close unclosed fences before the lexer sees them?

mvanhorn:
## Summary
- Fixes the animate plugin to include code/pre elements in its animation pass when the code fence is still open (incomplete)
- Adds `setAnimateCodeBlocks` method to `AnimatePlugin` interface so the Block component can toggle code animation per-block
- Once the code fence closes, animation reverts to default behavior (skip code blocks for syntax highlighting compatibility)

## Problem
When using `<Streamdown animated isAnimating>`, prose text streams with word-by-word fade-in animation. But the animate plugin unconditionally skipped all `code` and `pre` elements (via `SKIP_TAGS`), so code block content appeared instantly without any animation. This created a jarring visual discontinuity where code seemed to "buffer" and then dump all at once, while surrounding text streamed smoothly.

## Changes
- `packages/streamdown/lib/animate.ts`: Added `SKIP_TAGS_WITHOUT_CODE` set, made `hasSkipAncestor` state-aware, added `animateCodeBlocks` flag to render state, exposed `setAnimateCodeBlocks` on the plugin API
- `packages/streamdown/index.tsx`: Block component calls `animatePluginProp.setAnimateCodeBlocks(isIncomplete)` to enable code animation for blocks with unclosed fences

## Test plan
- [x] All 982 existing tests pass
- [x] Animate-specific tests pass (32 tests)
- [x] Incomplete code block tests pass (35 tests)
- [x] Biome linter passes with no issues
- [ ] Manual verification: stream a markdown response containing a fenced code block and confirm content appears incrementally with animation

Fixes #473

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/packages/streamdown/index.tsx
+++ b/packages/streamdown/index.tsx
@@ -347,6 +347,10 @@ export const Block = memo(
     if (animatePluginProp) {
       const prevCount = animatePluginProp.getLastRenderCharCount();
       animatePluginProp.setPrevContentLength(prevCount);
+      // When the block has an incomplete code fence, animate code block
+      // content incrementally instead of skipping it. This gives visual
+      // feedback that code is streaming in token-by-token.
+      animatePluginProp.setAnimateCodeBlocks(isIncomplete);
     }
 
     // Note: remend is already applied to the entire markdown before parsing into blocks
--- a/packages/streamdown/lib/animate.ts
+++ b/packages/streamdown/lib/animate.ts
@@ -12,6 +12,12 @@ export interface AnimatePlugin {
   name: "animate";
   rehypePlugin: Pluggable;
   /**
+   * Enable or disable animation inside code/pre blocks.
+   * When true, code block content will be animated incrementally
+   * during streaming (useful for incomplete/unclosed code fences).
+   */
+  setAnimateCodeBlocks: (enabled: boolean) => void;
+  /**
    * Set the number of HAST text characters from a previous render.
    * Characters up to this count will get duration=0ms, preventing
    * re-animation of already-visible content during streaming updates.
@@ -31,6 +37,7 @@ export interface AnimateOptions {
 const WHITESPACE_RE = /\s/;
 const WHITESPACE_ONLY_RE = /^\s+$/;
 const SKIP_TAGS = new Set(["code", "pre", "svg", "math", "annotation"]);
+const SKIP_TAGS_WITHOUT_CODE = new Set(["svg", "math", "annotation"]);
 
 const isElement = (node: unknown): node is Element =>
   typeof node === "object" &&
@@ -38,10 +45,15 @@ const isElement = (node: unknown): node is Element =>
   "type" in node &&
   (node as Element).type === "element";
 
-const hasSkipAncestor = (ancestors: Node[]): boolean =>
-  ancestors.some(
-    (ancestor) => isElement(ancestor) && SKIP_TAGS.has(ancestor.tagName)
+const hasSkipAncestor = (
+  ancestors: Node[],
+  animateCodeBlocks: boolean
+): boolean => {
+  const tags = animateCodeBlocks ? SKIP_TAGS_WITHOUT_CODE : SKIP_TAGS;
+  return ancestors.some(
+    (ancestor) => isElement(ancestor) && tags.has(ancestor.tagName)
   );
+};
 
 const splitByWord = (text: string): string[] => {
   const parts: string[] = [];
@@ -126,6 +138,7 @@ interface AnimateConfig {
  * object that setPrevContentLength / getLastRenderCharCount mutate.
  */
 interface AnimateRenderState {
+  animateCodeBlocks: boolean;
   lastRenderCharCount: number;
   prevContentLength: number;
 }
@@ -143,7 +156,7 @@ const processTextNode = (
     return;
   }
 
-  if (hasSkipAncestor(ancestors)) {
+  if (hasSkipAncestor(ancestors, renderState.animateCodeBlocks)) {
     return SKIP;
   }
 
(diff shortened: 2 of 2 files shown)


mvanhorn:
Friendly ping on this incremental code block rendering fix. Small change (+25/-4) and CI passes. Let me know if anything needs updating.

farnabaz:
@machielvdw Thanks for the PR, Could you update your PR and resolve conflicts?
Also I would love to have @Aventerica89 review and confirm over this after conflict resolved 🙂

machielvdw:
> @machielvdw Thanks for the PR, Could you update your PR and resolve conflicts? Also I would love to have @Aventerica89 review and confirm over this after conflict resolved 🙂

Hi @farnabaz. I think you meant to mention @mvanhorn

farnabaz:
@machielvdw My bad 😁 Thanks for replying 🙏

mvanhorn:
@farnabaz picking this back up. Before I push anything I want to check it is still wanted, because the answer changed while this sat.

It does not rebase. #594 already removed `code` from the animation skip list with the reasoning that inline code is layout-neutral, which covers part of what this PR was doing. And the `getLastRenderCharCount()` / `setPrevContentLength()` pair the hook relied on has been replaced by `commit()`, so the call site no longer exists in that form.

What is left of the original problem is narrower: while a fence is still open, `pre` is skipped, so streaming code appears in one dump while the prose around it fades in word by word. Fixing just that is not a two-line change any more. Animated word spans and Shiki tokens cannot both own the same subtree, so the open-fence path has to render plain animated content and hand over to the highlighted render once the fence closes. I have that working locally, with the docs updated and the suite green, but it is roughly 8 files rather than the 2 you last looked at.

So: still want it? I am happy to push the reworked version for review, trim it to something smaller if you would rather, or close this and let #594 stand if the remaining case is not worth that much surface area. Your call, I do not want to drop a much larger diff on you unannounced.

Also, no rush from my side, sorry this went quiet for so long.

farnabaz:
Thanks @mvanhorn for the reply and for taking the time to rework this against the current tree.

Animating incomplete code blocks does sound appealing, but my main concern is the highlight handoff. If we skip Shiki while the fence is open, long blocks would stream as plain text and then light up all at once when the fence closes. That feels like a worse UX than what we have today: keep pre on the animation skip list and keep highlighting during the stream, even if the content arrives more as a dump than word-by-word.

So my bias is to keep the current tradeoff (highlight over per-token code animation) unless the draft shows something better in practice. Happy to be wrong on that, if I haven’t considered something, tell me.

I also don’t want to decide this from the description alone. If you still have the working draft locally, I’d appreciate you pushing it so I can compare both states side by side before we commit either way. No rush on my side either.

Issue #478, opened 2026-03-25
Markdown syntax fails to render inside custom tags


hxj0x:
### Bug Description

Markdown syntax fails to render inside custom tags when content spans multiple lines

### Steps to Reproduce

```ts
import { Streamdown } from "streamdown";

// ❌ FAILS: "**bold**" renders as plain text, not bold.
// The content IS inside the component (unlike the header bug), but formatting is lost.
const markdown = "<ai-thinking>\n**bold**</ai-thinking>";

// ✅ WORKS: Single line bold works perfectly.
const markdownSingle = "<ai-thinking>**bold**</ai-thinking>";


function AiThinking(props) {
  return <div style={{ fontSize: 14, backgroundColor: "#3f3f3f3f" }}>{props.children}</div>;
}

export default function Demo() {
  return (
    <Streamdown
        allowedTags={{ "ai-thinking": [] }}
        components={{ "ai-thinking": AiThinking }}
    >
      {markdown}
    </Streamdown>
  );
}
```

### Expected Behavior

When content inside a custom tag contains newlines, the parser should recursively process the inner content as standard Markdown.

### Actual Behavior

Markdown syntax fails to render inside custom tags

### Code Sample

```tsx

```

### Streamdown Version

2.5.0

### React Version

18

### Node.js Version

22

### Operating System

None

PR #481, opened 2026-03-26
fix(custom-tags): restructure inline content with newlines to prevent ATX heading block splitting

### LINKED ISSUE #477: Custom tags prematurely close when followed by a newline and a Markdown header (e.g., ##)
### Bug Description

Custom components defined via allowedTags and components fail to wrap content correctly if the content inside the tag contains a newline followed by a Markdown header syntax (e.g., ## Title).
The parser appears to incorrectly terminate the custom tag scope upon encountering the newline + header combination, causing the header and subsequent content to be rendered outside the custom component.

### Steps to Reproduce

Here is a minimal reproduction example:
```ts

const markdown = "<ai-thinking> hi\n # yes</ai-thinking>"; 
// Expected: Both "hi" and "# yes" should be inside the AiThinking component.
// Actual: "hi" is inside, but "\n # yes" is rendered outside the component.

function AiThinking(props) {
  return <div style={{ fontSize: 14, backgroundColor: "#3f3f3f3f" }}>{props.children}</div>;
}

export default function Page() {
  return (
    <Streamdown
      allowedTags={{ "ai-thinking": [] }}
      components={{ "ai-thinking": AiThinking }}
    >
      {markdown}
    </Streamdown>
  );
}
```

### Expected Behavior

* The content <ai-thinking> hi\n # yes</ai-thinking> should be fully parsed within the ai-thinking tag.
* The AiThinking component should receive both hi and # yes (rendered as an h1) as children.
* Markdown syntax (like # yes) inside the custom tag should still be processed correctly even after a newline.

### Actual Behavior

* The parser seems to close the ai-thinking tag immediately after the first newline.
* The text \n # yes is rendered as plain text outside the gray background box defined in AiThinking.
* Note: Single line bold text like <ai-thinking>**bold**</ai-thinking> works fine. The issue specifically triggers with newlines.

### Streamdown Version

2.5.0

### React Version

18

### Node.js Version

22

### Browser(s)

_No response_

### Operating System

None

### Additional Context

_No response_

sleitor:
## Problem

Custom tags with multiline content are prematurely closed by the CommonMark parser when the content starts inline (on the same line as the opening tag) and contains a newline followed by a block-level element like an ATX heading.

**Minimal reproduction:**
```tsx
const markdown = '<ai-thinking> hi\n # yes</ai-thinking>';
// Expected: Both 'hi' and '# yes' inside the AiThinking component
// Actual: 'hi' inside, '# yes' rendered outside
```

The `marked` Lexer splits this as:
1. `paragraph`: `<ai-thinking> hi\n`
2. `heading`: ` # yes</ai-thinking>`

## Root Cause

`preprocess-custom-tags.ts` only restructures tags to block-level HTML when there are blank lines (`\n\n`) in the content. However, a single newline inside inline content is sufficient to introduce block-starting sequences (ATX headings, thematic breaks, list items, etc.) that break the parser.

Content that already starts on its own line (`\nContent\n`) is correctly parsed as a block-level HTML element by the Lexer — even with headings inside — because the opening tag is on its own line.

## Fix

Extend the restructuring condition to also apply when:
- Content is **inline** (doesn't start with `\n`), **AND**
- Content contains **any newline**

Tags where content already starts on its own line (block-level) are unaffected. Single-line tags without newlines are also unaffected.

Fixes #477
Related: #478 (same root cause)

### DIFF
--- a/packages/streamdown/__tests__/preprocess-custom-tags.test.ts
+++ b/packages/streamdown/__tests__/preprocess-custom-tags.test.ts
@@ -51,9 +51,23 @@ describe("preprocessCustomTags", () => {
     expect(preprocessCustomTags(md, ["custom"])).toBe(md);
   });
 
-  it("should not modify content without blank lines", () => {
+  it("should not modify content without blank lines (no newline)", () => {
+    // Single-line content without any newlines is fine as inline HTML
     const md = "<custom>Hello World</custom>";
     expect(preprocessCustomTags(md, ["custom"])).toBe(md);
+  });
+
+  it("should restructure inline content with a newline followed by ATX heading", () => {
+    // A heading inside inline content would prematurely terminate the custom tag
+    const md = "<ai-thinking> hi\n # yes</ai-thinking>";
+    const result = preprocessCustomTags(md, ["ai-thinking"]);
+    expect(result).toBe("<ai-thinking>\n hi\n # yes\n</ai-thinking>\n\n");
+  });
+
+  it("should restructure inline content with any newline (prevents block element splitting)", () => {
+    const md = "<custom>hello\nworld</custom>";
+    const result = preprocessCustomTags(md, ["custom"]);
+    expect(result).toBe("<custom>\nhello\nworld\n</custom>\n\n");
   });
 
   it("should not modify tags where content already starts on own line without blank lines", () => {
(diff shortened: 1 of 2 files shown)


farnabaz:
close in favor of #484

Issue #485, opened 2026-03-30
streamdown-mermaid: toolbar icon button contains unlabeled SVG and triggers accessibility warning


simon5057:
### Bug Description

Toolbar buttons in the Mermaid integration contain inline <svg> icons that are exposed to the accessibility tree, triggering the warning: "Content with images must be labeled." The buttons themselves have labels, but the SVGs inside lack proper aria-hidden attributes.

<img width="466" height="378" alt="Image" src="https://github.com/user-attachments/assets/d65bd210-ce6f-4ce4-bf16-5d4cc2dd6e06" />

<img width="410" height="474" alt="Image" src="https://github.com/user-attachments/assets/c5df04a5-7c6f-44c6-b3da-256c19b401d2" />

### Steps to Reproduce

1. Render Mermaid content via streamdown/mermaid
2. Inspect toolbar buttons adjacent to the diagram
3. Run accessibility audit(Firefox Accessibility Inspector, axe DevTools)

### Expected Behavior

Decorative SVG icons should be hidden from the accessibility tree.

### Actual Behavior

The toolbar buttons around Mermaid-rendered content have no text alternative for screen reader users.

### Code Sample

```tsx

```

### Streamdown Version

2.5.0

### React Version

19.0.0

### Operating System

None

PR #486, opened 2026-03-30
fix(icons): add aria-hidden attribute to all icon components for accessibility(#485)

simon5057:
### Problem

Toolbar buttons in the Mermaid integration (and other components) contain inline SVG icons that are exposed to the accessibility tree, triggering the warning: **"Content with images must be labeled"** in accessibility audits (Firefox Accessibility Inspector, axe DevTools). While the parent buttons have `title` attributes providing accessible names, the SVGs themselves lacked proper `aria-hidden` attributes to mark them as decorative.

### Solution

Added `aria-hidden="true"` to all 10 SVG icon components in `packages/streamdown/lib/icons.tsx`:

- `CheckIcon`
- `CopyIcon`
- `DownloadIcon`
- `Loader2Icon`
- `Maximize2Icon`
- `RotateCcwIcon`
- `XIcon`
- `ExternalLinkIcon`
- `ZoomInIcon`
- `ZoomOutIcon`

This hides the decorative SVGs from the accessibility tree. Screen readers will rely on the parent button `title` attributes instead, which already provide meaningful labels (e.g., "Zoom in", "Copy Code", "Download").

### Testing

- Verified with accessibility audit tools that the "Content with images must be labeled" warning is resolved.

### DIFF
--- a/.changeset/fix-icons-aria-hidden.md
+++ b/.changeset/fix-icons-aria-hidden.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Add `aria-hidden="true"` to decorative SVG icon components so screen readers rely on parent button labels instead of unlabeled icons.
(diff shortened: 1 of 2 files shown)


farnabaz (approved):
Thanks

PR #500, opened 2026-04-09
feat: add configurable portal target

### LINKED ISSUE #499: feat(mermaid): configurable portal target for Mermaid fullscreen (micro-frontends & prefixed CSS)
### Problem

Streamdown is often embedded in **micro-frontend** setups: the markdown renderer runs inside a **child app** or **isolated subtree** of a larger shell, not as the only root on the page.

Those hosts frequently rely on:

- **Scoped CSS** — e.g. a wrapper class around the sub-app so design tokens and utilities apply only there.
- **Prefixed Tailwind** — e.g. Tailwind v4 **`prefix()`** or similar, so generated classes are **`tw:flex`**, **`tw:fixed`**, etc., and only match elements **under** the scanned scope.

Mermaid **fullscreen** is implemented with **`createPortal`**, typically to **`document.body`**. The fullscreen layer is then **outside** the micro-frontend’s DOM subtree. As a result:

- **Prefixed / scoped utility classes** from the host may **not apply** to nodes under `document.body`, so layout, positioning, and **`fixed` / `z-index`** behavior can be wrong or missing.
- **Stacking** can break: the overlay may render **under** shell chrome (headers, footers, sidebars) that live in a different part of the tree with higher **z-index**.

This is the same **portal vs in-tree** class of issue as elsewhere in Streamdown (e.g. table fullscreen and DOM context — see #468 / PR #470).

Separately, #490 proposes **host callbacks** to **replace** download/fullscreen (e.g. desktop webviews). That addresses hosts that want to **own** the action. Many embedded apps instead need to **keep** the default fullscreen UI but have it mount **inside** their subtree so **prefix + scope** still apply.

### Proposed solution

Add an optional, **backward-compatible** API to control **where** the Mermaid fullscreen portal mounts:

- **Name (example):** `fullscreenPortalContainer` or `mermaid.fullscreen.portalContainer`
- **Type:** `HTMLElement | null` or `() => HTMLElement | null` (match idioms you use elsewhere).
- **Behavior:** When set, Mermaid fullscreen (and any overlay that should share the same stacking context) portals into this element instead of **`document.body`**.
- **Default:** **`document.body`** — unchanged for existing consumers.

**Docs:** Short note for **micro-frontends** and **prefixed / scoped Tailwind**: provide a **portal root element inside** the same wrapper as `<Streamdown />` and pass it here so fullscreen stays in the **same DOM and CSS scope** as the rest of the sub-app.

### Alternatives considered

- **Relying only on global CSS / safelist:** Duplicates tokens, fights prefixing, and drifts from the shell’s theme.
- **Replacing the entire Mermaid fence via a custom renderer:** Too heavy for fixing portal placement alone.

### Related issues

- #468 / #470 — portal and DOM context (tables).
- #490 — host hooks to skip default fullscreen; complementary to a **portal root** for hosts that keep default UI.

### Contribution

Happy to help with a PR (API surface + tests) once maintainers agree on naming and placement on `Streamdown` vs the Mermaid plugin.

---

**Pull request:** https://github.com/vercel/streamdown/pull/500

slavco86:
## Summary

Adds a top-level **`portal`** prop to configure the container used by Streamdown's built-in overlays. This keeps overlays inside a host subtree when using micro-frontends, scoped CSS, prefixed Tailwind utilities, or a custom stacking context.

## Changes

- Add `portal?: HTMLElement | null | (() => HTMLElement | null)` to `Streamdown`.
- Use the configured target for:
  - Mermaid fullscreen
  - Table fullscreen
  - The built-in link safety modal
- Keep `document.body` as the backward-compatible fallback.
- Resolve getter targets only when an overlay opens, preserving SSR safety.
- Recompute the memoized context when `portal` changes.
- Document usage and add a patch changeset.

A custom `linkSafety.renderModal` remains host-controlled and therefore controls its own placement.

## Testing

- `pnpm build:packages`
- `pnpm build --filter website`
- `pnpm test` — 1,144 tests passed
- `pnpm check`
- Focused portal and overlay tests — 53 tests passed

Fixes #499

### DIFF
--- a/packages/streamdown/index.tsx
+++ b/packages/streamdown/index.tsx
@@ -180,6 +180,12 @@ export interface MermaidErrorComponentProps {
 export interface MermaidOptions {
   config?: MermaidConfig;
   errorComponent?: React.ComponentType<MermaidErrorComponentProps>;
+  /**
+   * DOM node for the Mermaid fullscreen overlay (`createPortal` target).
+   * Use a container inside your app root when using scoped CSS or prefixed Tailwind so `fixed` / z-index utilities apply.
+   * @default document.body
+   */
+  fullscreenPortalContainer?: HTMLElement | null | (() => HTMLElement | null);
 }
 
 export type AllowedTags = Record<string, string[]>;
@@ -859,6 +865,7 @@ export const Streamdown = memo(
     prevProps.plugins === nextProps.plugins &&
     prevProps.className === nextProps.className &&
     prevProps.linkSafety === nextProps.linkSafety &&
+    prevProps.mermaid === nextProps.mermaid &&
     prevProps.lineNumbers === nextProps.lineNumbers &&
     prevProps.normalizeHtmlIndentation === nextProps.normalizeHtmlIndentation &&
     prevProps.literalTagContent === nextProps.literalTagContent &&
--- a/packages/streamdown/index.tsx
+++ b/packages/streamdown/index.tsx
@@ -192,6 +192,7 @@ export interface MermaidOptions {
 }
 
 export type AllowedTags = Record<string, string[]>;
+export type PortalTarget = HTMLElement | null | (() => HTMLElement | null);
 
 export type StreamdownProps = Options & {
   mode?: "static" | "streaming";
@@ -230,6 +231,11 @@ export type StreamdownProps = Options & {
   plugins?: PluginConfig;
   remend?: RemendOptions;
   linkSafety?: LinkSafetyConfig;
+  /**
+   * DOM node for Streamdown overlays, or a function returning one.
+   * Defaults to `document.body`.
+   */
+  portal?: PortalTarget;
   /** Custom tags to allow through sanitization with their permitted attributes */
   allowedTags?: AllowedTags;
   /**
@@ -320,6 +326,7 @@ export interface StreamdownContextType {
   linkSafety?: LinkSafetyConfig;
   mermaid?: MermaidOptions;
   mode: "static" | "streaming";
+  portal?: PortalTarget;
   shikiTheme: [ThemeInput, ThemeInput];
   /** Max height for tables. @default 300 */
   tableMaxHeight: number | string;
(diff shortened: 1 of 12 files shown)


farnabaz (changes requested):
Thanks for the PR @slavco86,
I like the idea of having portal container option, but we also have similar feature for link and table modals. It make no sense to have this feature for mermaid but not for others.

Instead of `mermaid.fullscreenPortalContainer` we can introduce `portal` directly as Streamdown prop and use it in all of modal.

Are you happy to update your PR to create this prop?

slavco86:
@farnabaz , thanks for picking this up and commenting!

Yes, absolutely - that makes perfect sense. 

Leave it with me

farnabaz:
@slavco86 one other thing, I appreciate if you sign all your commits, new and old ones. It is organization policy, all commits should be signed in order to merge a PR. (otherwise PR is blocked)

https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

slavco86:
@farnabaz Updated as requested:

- Replaced the Mermaid-specific option with a top-level `portal` prop.
- Applied it to Mermaid fullscreen, table fullscreen, and the built-in link safety modal (all current `createPortal` call sites).
- Rebased the PR onto current `main`; it is now conflict-free.
- Replaced the old unsigned history with one GitHub-verified signed commit.
- Added integration tests, SSR-lazy-resolution coverage, docs, and a patch changeset.

The full package build, website build, 1,144-test suite, and lint/format checks pass locally. Ready for another review when convenient.

packages/streamdown/index.tsx
```diff
@@ -978,6 +989,7 @@ export const Streamdown = memo(
     prevProps.plugins === nextProps.plugins &&
     prevProps.className === nextProps.className &&
     prevProps.linkSafety === nextProps.linkSafety &&
+    prevProps.portal === nextProps.portal &&
```
farnabaz: @slavco86  If we pass function to portal, when other props change, component will re-render because portal will receive new instance.
I think it is good to remove this and treat portal as initializing props like other function props (`urlTransform` and `allowElement`)

slavco86: @farnabaz , good point! Thanks. Updated

farnabaz (approved):
Thanks

PR #504, opened 2026-04-10
feat: render loading placeholder for incomplete streaming images

### LINKED ISSUE #503: Streaming support for partial images (similar to incomplete link handling)
### Problem Statement

When streaming markdown, ‎`streamdown` currently removes incomplete images (e.g. ‎`alt instead of rendering a placeholder. This works fine for final output, but in streaming UIs it introduces a noticeable delay: nothing is shown until the full image URL is emitted and the markdown becomes valid. In contrast, links are handled more gracefully: incomplete links are replaced with a placeholder URL (‎`streamdown:incomplete-link`), so the user can already see the link text while the model is still streaming.

For image-heavy content, this behavior makes the streaming experience feel less responsive compared to how links are handled.

### Proposed Solution

Introduce a streaming-friendly image handling mode, similar to incomplete link handling. For example:

- When an image is incomplete (unterminated URL or closing ‎`)` missing), render a placeholder image or a “loading” image state instead of dropping the node.

- Optionally provide a configurable placeholder strategy (e.g. built-in default, custom React component, or callback).

- Once the full image URL has been streamed, update the rendered node in place to the real image.

The goal is that the user can see that “an image is coming” immediately, even before the markdown for the image is fully complete.

### Alternatives Considered

- Keeping the current behavior and waiting until the full image URL is streamed. This keeps the implementation simple but gives a worse UX for streaming, especially when models emit long or slow URLs.

- Post-processing the model output on the host side to inject a custom placeholder when ‎`![` is detected without a closing ‎`)`. This is possible but reimplements logic that ‎`streamdown` (and ‎`remend`) already understand, and is easy to get wrong for edge cases.

- Using text-only placeholders in the original markdown (e.g. “Loading image…”), then replacing them later. This requires changing prompts / model behavior and doesn’t leverage the existing markdown/image pipeline.

### Use Case

```tsx
In a React app that streams AI responses using ‎`streamdown`, we often generate content with multiple images (e.g. visual step‑by‑step guides, design reviews, or diagrams). As the model streams:

- Links are visible right away thanks to incomplete link handling.

- Images, however, simply don’t appear until the entire image markdown is complete, which can be several seconds later.
```

### Priority

Nice to have

### Contribution

### Additional Context

_No response_

sleitor:
## Summary

Fixes #503 — incomplete images during streaming now render an animated skeleton placeholder instead of being silently removed. This mirrors the existing behavior for incomplete links.

## What changed

****
- `link-image-handler.ts`: Both `handleIncompleteUrl()` and `handleIncompleteText()` now emit `alt` for incomplete images instead of stripping them entirely
- `index.ts`: Updated `earlyReturn` check to also short-circuit on `streamdown:incomplete-image`; fixed JSDoc for `images` option
- Tests updated: removal expectations → placeholder expectations

**`packages/streamdown`**
- `lib/image.tsx`: `ImageComponent` detects `src === 'streamdown:incomplete-image'` and renders an `animate-pulse` skeleton div with `data-incomplete="true"`
- Tests added for the new placeholder behavior

**Changeset**: `remend@patch` + `streamdown@patch`

## Before / After

**Before**: `logo

**After**: `![logo](./assets/log` → animated grey skeleton placeholder shown until the stream completes

## Test coverage

- 356 remend tests ✅
- 964 streamdown tests ✅

All existing tests updated; 4 new tests added for the placeholder behavior.


farnabaz (changes requested):
Thanks for the PR @sleitor 
Do you mind removing line numbers commit from this PR?

Also I do think `streamdown` schema will be blocked by `rehype-harden` package we need to change placeholder to another text. maybe even an 1x1 pixel placeholder
```
data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="1" height="1"/>
// or equivalent png
```

sleitor:
Thanks for the review, @farnabaz!

- Removed the unrelated `lineNumbers` commit that got mixed into this branch (rebased, now only the placeholder feature + this fix remain).
- Replaced the `streamdown:incomplete-image` custom-scheme marker with a 1x1 transparent PNG `data:` URI (`INCOMPLETE_IMAGE_PLACEHOLDER`, exported from `remend`). This should survive strict URL-sanitizing schemas like `rehype-harden`, since only `http(s)`/`data` schemes are typically allowlisted — a bare custom scheme string would likely get stripped before it reaches the `ImageComponent`'s `src` check.

All 356 remend + 951 streamdown tests pass, lint is clean. Let me know if you'd like a different data URI (e.g. transparent GIF/SVG) or any other adjustments.

farnabaz:
Thanks @sleitor,
The more I think, the more get to this idea that your initial work on `src` was right, having empty 1x1 image in the src make no sense. `streamdown:incomplete-image` makes it smaller and easier to detect.
I just updated you PR to put back `streamdown:incomplete-image`  and add `streamdown` protocol in `src` accepted protocols.

farnabaz (approved):

PR #507, opened 2026-04-16
fix(mermaid): fullscreen overlay accessibility and add stable data-streamdown attribute

aradhyacp:
## Description

This PR fixes accessibility and consistency issues in the mermaid fullscreen overlay by aligning it with the existing table fullscreen implementation. It also introduces a stable attribute for external styling and tooling.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Related Issues

Fixes #506
Closes #506
Related to #506

## Changes Made

- Added `data-streamdown="mermaid-fullscreen"` to the fullscreen overlay for stable CSS targeting
- Updated overlay semantics from `role="button"` to `role="dialog"`
- Added `aria-modal="true"` to improve screen reader modal behavior
- Aligned mermaid fullscreen overlay behavior with table fullscreen implementation

## Testing

- [x] All existing tests pass
- [ ] Added new tests for the changes
- [x] Manually tested the changes

### Test Coverage

N/A (UI-only accessibility and structural change)

## Checklist

- [x] My code follows the project's code style
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I have created a changeset (`pnpm changeset`)

## Changeset

- [x] I have created a changeset for these changes

## Additional Notes

This change brings parity between mermaid and table fullscreen overlays and improves accessibility compliance for modal behavior in screen readers.

PR #508, opened 2026-04-16
fix(mermaid): add data-streamdown, aria-modal, and correct role to fullscreen overlay

### LINKED ISSUE #506: Mermaid fullscreen overlay missing data-streamdown attribute and accessibility attributes
## Summary

The mermaid fullscreen overlay (rendered via `createPortal` to `document.body`) is missing the `data-streamdown` attribute, `aria-modal`, and uses `role="button"` instead of `role="dialog"`. The table fullscreen overlay correctly has all three.

## Current behavior (v2.4.0 / v2.5.0)

**Table fullscreen overlay** (correct):
```jsx
createPortal(jsx("div", {
  "aria-label": d.viewFullscreen,
  "aria-modal": "true",
  className: "fixed inset-0 z-50 flex flex-col bg-background",
  "data-streamdown": "table-fullscreen",
  role: "dialog",
  // ...
}), document.body)
```

**Mermaid fullscreen overlay** (missing attributes):
```jsx
createPortal(jsxs("div", {
  className: "fixed inset-0 z-50 flex items-center justify-center bg-background/95 backdrop-blur-sm",
  role: "button",  // should be "dialog"
  tabIndex: 0,
  // missing: data-streamdown, aria-modal
}), document.body)
```

## Why this matters

1. **CSS targeting**: Without `data-streamdown`, consumers cannot target the mermaid fullscreen overlay with stable CSS selectors. The only option is fragile class-based selectors like `.bg-background\/95.fixed.inset-0` which break on any class rename.

2. **Accessibility**: The overlay is a modal dialog (full-viewport, traps focus, closes on Escape) but uses `role="button"` and lacks `aria-modal="true"`. Screen readers cannot identify it as a modal.

3. **Inconsistency**: Table fullscreen has `data-streamdown="table-fullscreen"`, `aria-modal="true"`, and `role="dialog"`. The mermaid fullscreen should follow the same pattern.

## Discovered via

We hit this while fixing a z-index layering issue — our chat drawer uses `z-index: 200`, and the mermaid fullscreen overlay at `z-50` was hidden behind it on mobile. Without `data-streamdown`, we had to use fragile class selectors to elevate the z-index. We're currently working around this with a `pnpm patch`.

## Proposed fix

Add to the mermaid fullscreen overlay element:
- `"data-streamdown": "mermaid-fullscreen"`
- `"aria-modal": "true"`
- `role: "dialog"` (instead of `role: "button"`)

This matches the existing table fullscreen pattern and enables stable CSS targeting + correct accessibility semantics.

sleitor:
- Add `data-streamdown="mermaid-fullscreen"` for stable CSS targeting
- Change `role="button"` to `role="dialog"` (correct semantics for a modal overlay)
- Add `aria-modal="true"` for screen readers
- Add `aria-label` for accessible name
- Remove `tabIndex={0}` (not needed on dialog element)
- Update biome-ignore comment to match table fullscreen pattern

Closes #506

### DIFF
--- a/.changeset/fix-mermaid-fullscreen-a11y.md
+++ b/.changeset/fix-mermaid-fullscreen-a11y.md
@@ -1,0 +1,7 @@
+---
+"streamdown": patch
+---
+
+fix(mermaid): add data-streamdown, aria-modal, and correct role to fullscreen overlay
+
+The Mermaid fullscreen overlay was missing the `data-streamdown="mermaid-fullscreen"` attribute and using `role="button"` instead of `role="dialog"`, and was missing `aria-modal="true"`. This matches the table fullscreen overlay pattern and enables stable CSS targeting and correct accessibility semantics.
--- a/packages/streamdown/lib/mermaid/fullscreen-button.tsx
+++ b/packages/streamdown/lib/mermaid/fullscreen-button.tsx
@@ -93,19 +93,21 @@ export const MermaidFullscreenButton = ({
 
       {isFullscreen
         ? createPortal(
-            // biome-ignore lint/a11y/useSemanticElements: "div is used as a backdrop overlay, not a button"
+            // biome-ignore lint/a11y/noNoninteractiveElementInteractions: "dialog overlay needs click-to-dismiss"
             <div
+              aria-label={t.viewFullscreen}
+              aria-modal="true"
               className={cn(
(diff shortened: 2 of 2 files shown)


farnabaz:
Thanks for the PR, closing this in favor of #507

PR #517, opened 2026-04-30
fix(mermaid): use XMLSerializer to produce valid SVG/XML on download

### LINKED ISSUE #516: Mermaid download produces invalid SVG / fails for PNG when chart contains <br />
### Bug Description

Downloading a Mermaid flowchart chart that contains `<br/>` in a label produces a broken file:
- SVG opens with an XML parse error in the browser and is incomplete (stops rendering after `br`)
- PNG is not downloaded without errors in console

In-page preview renders correctly, bug only affects the download.

Affected chart types:
* `flowchart`
* `stateDiagram-v2`

`sequence` and `classDiagram` are not affectd.

### Steps to Reproduce

1. Go to playground: https://streamdown.ai/playground
2. Use this example:
     ````markdown
     ```mermaid
     flowchart TD
       A["line 1<br/>line 2"] --> B["B"]
     ```
     ````
2. Click download button
3. Select "PNG"
4. Select "SVG"

Image

Example for state diagram:
````markdown
```mermaid
stateDiagram-v2                                                          
  A --> B: x<br/>y
```
````

### Expected Behavior

* Download PNG should work
* Downloaded SVG should open without errors and be rendered in full

### Actual Behavior

* PNG - nothing happens
* SVG - file is saved, but opens with error in browser, not fully rendered

Image

### Code Sample

```tsx
import { mermaid } from "@streamdown/mermaid";

const md = `
  \`\`\`mermaid                                                               
  flowchart TD                                           
      A["x<br/>y"]
  \`\`\`
`;

export default () => <Streamdown plugins={{ mermaid }}>{md}</Streamdown>;
```

### Streamdown Version

2.5.0 (playground), 1.6.11 (first encountered in my project)

### React Version

18.3.1 (my project)

### Node.js Version

v24.6.0

### Browser(s)

Chrome

### Operating System

macOS

### Additional Context

Core problem: `mermaid.render()` returns HTML, not XML. The problem with `<br />` is that it is normalized to `<br>` (without closing slash) in browser. So then when it is saved to SVG as is, it comes as invalid XML.

Possible fixes:
* Fixed it in my project with yarn patch by doing `svg.replace(/<br>/g, "<br/>")` before writing to SVG.
* `@mermaid-js/mermaid-cli` uses `XMLSerializer.serializeToString`, but it requires DOM element as input. See https://github.com/mermaid-js/mermaid-cli/blob/f87dcd0c9049a58ba75eda6d6dc890e2285be850/src/index.js#L352-L360 

```js
    if (outputFormat === 'svg') {
      const svgXML = await page.$eval('svg', (svg) => {
        // SVG might have HTML <foreignObject> that are not valid XML
        // E.g. <br> must be replaced with <br/>
        // Luckily the DOM Web API has the XMLSerializer for this
        // eslint-disable-next-line no-undef
        const xmlSerializer = new XMLSerializer()
        return xmlSerializer.serializeToString(svg)
      })
```

sleitor:
## Problem

Mermaid renders diagrams via `innerHTML`, which allows the browser to normalize elements like `<br />` to their HTML void form (`<br>`). Saving that raw string as an SVG file fails because SVG/XML parsers reject the non-self-closing element, and PNG conversion via canvas also fails since the `` cannot load invalid XML SVG.

Reported in #516.

## Fix

Add `sanitizeSvgForExport()` in `lib/mermaid/utils.ts` that:
1. Parses the raw SVG string into the DOM via a temporary `<div>.innerHTML`
2. Re-serializes the `<svg>` element with `XMLSerializer.serializeToString()`, producing a well-formed XML document
3. Falls back to the original string if running outside a browser or if parsing fails

Both the SVG download path and the PNG conversion path in `download-button.tsx` now sanitize the SVG string before use.

## Tests

- Updated existing SVG/PNG download tests to accept XMLSerializer output (which adds the XML namespace attribute)
- Added new tests for the `<br>` element case in both `mermaid-download.test.tsx` and `mermaid-utils.test.ts`
- All 988 tests pass

Fixes #516

### DIFF
--- a/.changeset/fix-mermaid-svg-serialization.md
+++ b/.changeset/fix-mermaid-svg-serialization.md
@@ -1,0 +1,17 @@
+---
+"streamdown": patch
+---
+
+fix(mermaid): use XMLSerializer to produce valid SVG/XML on download
+
+Mermaid renders diagrams via `innerHTML`, which allows the browser to
+normalize elements like `<br />` to their HTML void form (`<br>`).
+Saving that raw string as an SVG file (or converting it to PNG via a
+canvas) fails because SVG/XML parsers reject the non-self-closing element.
+
+Add `sanitizeSvgForExport` which re-parses the SVG string through a
+temporary DOM node and re-serializes it with `XMLSerializer.serializeToString`,
+producing a valid XML document.  Both the SVG download path and the PNG
+conversion path now sanitize the SVG before use.
+
+Fixes #516
--- a/packages/streamdown/__tests__/mermaid-download.test.tsx
+++ b/packages/streamdown/__tests__/mermaid-download.test.tsx
@@ -17,6 +17,10 @@ vi.mock("../lib/mermaid/utils", () => ({
   svgToPngBlob: vi
     .fn()
     .mockResolvedValue(new Blob(["png"], { type: "image/png" })),
+  // Pass through – sanitize is a no-op in the test environment since the SVG
+  // strings used in tests are already valid; the real behaviour is covered by
+  // the sanitizeSvgForExport-specific tests in mermaid-utils.test.ts.
+  sanitizeSvgForExport: vi.fn((svg: string) => svg),
 }));
 
 describe("MermaidDownloadDropdown", () => {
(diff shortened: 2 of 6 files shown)


farnabaz:
Thanks for the PR,
I saw this after #526, which tackled same issue. Closing in favor of #526.

PR #518, opened 2026-05-02
fix(streamdown): correct CSS bracket syntax in code-block bg classes

Qiuhang817385:
## Summary
Fix CSS bracket syntax errors in `code-block/body.tsx` where Tailwind classes for background color were missing closing brackets, causing invalid class names to be generated.

- `bg-[var(--sdm-bg,inherit]` → `bg-[var(--sdm-bg,inherit)]`
- `dark:bg-[var(--shiki-dark-bg,var(--sdm-bg,inherit)]` → `dark:bg-[var(--shiki-dark-bg,var(--sdm-bg,inherit))]`

## Test plan
- [x] Verify dark mode background fallback works correctly when Shiki does not provide `--shiki-dark-bg`
- [x] Verify light mode background still works (backed by inline CSS variables)

### DIFF
--- a/packages/streamdown/lib/code-block/body.tsx
+++ b/packages/streamdown/lib/code-block/body.tsx
@@ -94,8 +94,8 @@ export const CodeBlockBody = memo(
         <pre
           className={cn(
             className,
-            "bg-[var(--sdm-bg,inherit]",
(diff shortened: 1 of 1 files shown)


farnabaz (approved):
Thanks

Issue #520, opened 2026-05-16
verify-changesets workflow fails and missing changesets are not guided


aradhyacp:
## Changeset workflow fails for fork contributors + missing changeset guidance

Hi team,

While creating PRs and reviewing PRs from other contributors, I noticed a couple of recurring issues related to the changeset workflow. Initially I did not pay much attention to it, but after seeing it repeatedly across multiple PRs, I thought it would be good to bring it into notice.

---

### 1. Workflow fails for fork PRs when a changeset is included

Whenever a PR from a fork includes a `.changeset/*.md` file, the `Verify Changesets` workflow fails immediately with:

```txt
Error: Input required and not supplied: app-id
````

From what I understand, this happens before the actual validation runs and seems related to the GitHub App secrets not being available for fork-based PR workflows.

Because of this, contributors who correctly include a changeset end up getting a failing workflow.

---

### 2. Missing changesets are not enforced or guided

Changesets are mentioned as required, but currently PRs without a changeset do not trigger any workflow or guidance.

Maybe instead of silently passing, the workflow could leave a small friendly reminder comment like:

> "Dear developer, please attach a changeset — it helps us a lot with release management."

This could help contributors understand the expectation without confusion.

---

Just wanted to report this since it seems to affect contributor experience repeatedly across PRs.

Thanks

### Steps to Reproduce

1. Fork the repository and create a new branch
2. Add a `.changeset/*.md` file in the PR
3. Open a PR from the forked repository to `vercel/streamdown:main`
4. Observe the `Verify Changesets` workflow failure:

```txt
Run actions/create-github-app-token@v1
/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:42555
  throw new Error("Input required and not supplied: app-id");

Error: Input required and not supplied: app-id
    at Object.<anonymous> (/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:42555:9)
    at Module._compile (node:internal/modules/cjs/loader:1521:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1623:10)
    at Module.load (node:internal/modules/cjs/loader:1266:32)
    at Module._load (node:internal/modules/cjs/loader:1091:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:164:12)
    at node:internal/main/run_main_module:28:49

Node.js v20.20.2
```

5. Create another PR without a changeset file
[...]

PR #525, opened 2026-05-20
fix: avoid app token in changeset verification

omribz156:
Summary:
- removed the GitHub App token generation from the changeset verification workflow
- let `actions/checkout` use the default read-only token so fork PRs do not need repository secrets
- removed the now-unused `GITHUB_TOKEN` env from the verifier step

Verification:
- `node --check .github/workflows/actions/verify-changesets/index.js`
- direct `verifyChangesets()` smoke test with a valid patch changeset
- `git diff --check -- .github/workflows/verify-changesets.yml .github/workflows/actions/verify-changesets/index.js`
- `actionlint .github/workflows/verify-changesets.yml`

This keeps the workflow on `pull_request` instead of moving to `pull_request_target`, so fork PR validation remains unprivileged. It addresses the fork-secret failure path from #520; I left the separate missing-changeset guidance behavior alone because that likely needs a maintainer preference on whether to comment, warn, or enforce.

Implemented with Codex assistance, with the workflow change kept focused and manually reviewed.

### DIFF
--- a/.github/workflows/verify-changesets.yml
+++ b/.github/workflows/verify-changesets.yml
@@ -21,17 +21,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
-      - name: Generate GitHub App Token
-        id: generate-token
-        uses: actions/create-github-app-token@v1
-        with:
-          app-id: ${{ secrets.STREAMDOWN_APP_ID }}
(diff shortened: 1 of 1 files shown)


farnabaz:
Thanks for the PR, sadly I didn't saw your PR sooner and I did already remove this part.
Closing this because it is not valid anymore.

PR #529, opened 2026-05-31
fix(link-modal): render link safety modal in a portal to hydration error

osipov-anton:
## Summary

When `linkSafety` is enabled, the link component renders a `<button>` alongside the `LinkSafetyModal`. The modal's root is a `<div>` rendered inline, but markdown links live inside paragraph `<p>` elements — so the modal `<div>` ends up as a descendant of `<p>`.

This produces the React hydration error:

> In HTML, `<div>` cannot be a descendant of `<p>`. This will cause a hydration error.

### Fix

Render the default `LinkSafetyModal` through `createPortal` to `document.body`, so the modal `<div>` is no longer nested inside the paragraph. The inline `<button>` (valid inside `<p>`) stays where it is. The component also guards against SSR (`typeof document === "undefined"`).

## Test plan

- [x] `pnpm --filter streamdown test` — all 982 tests pass
- [x] Updated `link-modal-keyboard`, `link-safety`, and `translations` tests to query the portaled DOM (`document` / `document.body`) instead of the render `container`
- [ ] Manually verify no hydration warning when an external link appears inside a paragraph with link safety enabled

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/fix-link-modal-portal.md
+++ b/.changeset/fix-link-modal-portal.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Render the link safety modal through a portal to `document.body` so it is no longer nested inside the paragraph's `<p>` element. This fixes the React hydration error "In HTML, `<div>` cannot be a descendant of `<p>`" that occurred when a link with link safety enabled appeared inside a paragraph.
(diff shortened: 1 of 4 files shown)


farnabaz (approved):
Thanks 👍

Issue #185, opened 2025-10-15
Plain CommonMark/GFM is incomaptible with Japanese/Chinese + bold output


tats-u:
### Bug Description

Plain CommonMark/GFM has a specification bug that is incompatible with Chinese/Japanese Markdown output using bold text:

<img width="804" height="250" alt="Image" src="https://github.com/user-attachments/assets/b56346c6-a34b-410c-a8a9-c15b125f96a9" />

↑ChatGPT + Japanese response

Humans can take care of `**` around punctuation, but It is much more difficult to get LLMs to pay attention to that. It is the best to modify the Markdown specification itself to eliminate that pitfall for LLMs.

See https://github.com/tats-u/markdown-cjk-friendly for the details.

Please include additional remark plugin(s) to deal with this bug or at lease add a note to the documentation.

### Steps to Reproduce

Parse the following Markdown content and render it in a GFM-compliant Markdown parser including streamdown:

```md
**この文は太字になりません（This sentence will not be bolded）。**この文のせいで（It is due to this sentence）。
```

In real productions, this is likely to occur when an LLM generates the output in the following situations: 

- Japanese or Chinese
- It tries to emphasize phrases that are surrounded by or end with ideographic parenthesis or brackets

### Expected Behavior

<div lang=ja>

> <strong>この文は太字になりません（This sentence will not be bolded）。</strong>この文のせいで（It is due to this sentence）。

</div>

↑I used a raw HTML tag `<strong>` here (see below for the reason)

### Actual Behavior

<div lang=ja>

> **この文は太字になりません（This sentence will not be bolded）。**この文のせいで（It is due to this sentence）。

</div>

↑ I pasted the Markdown source as is here. GitHub's Markdown parser fails to parse `**`, too!

### Code Sample

```tsx
import { Streamdown } from "streamdown";

export default function App() {
  const markdown = "**この文は太字になりません（This code is not bolded）。**この文のせいで（It is due to this sentense）。";

  return <Streamdown>{markdown}</Streamdown>;
}
```

### Streamdown Version

1.4.0

### React Version

This is not concerned with React's version.

### Node.js Version

This is not concerned with Node's version.

### Operating System

None

### Additional Context

- https://github.com/lobehub/lobe-chat
- https://github.com/lobehub/lobe-ui
- https://github.com/CherryHQ/cherry-studio

These managed to get over this bug thanks to my plugin [remark-cjk-friendly](https://npmjs.com/package/remark-cjk-friendly).

[...]

Issue #140, opened 2025-09-14
Maximum update depth exceeded.


Nishchit14:
### Bug Description

After implemnting ai-elements I am facing this issue while generating the longer responses. It is happening when stream is running more than 30sec.

```md
Maximum update depth exceeded. 
This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. 
React limits the number of nested updates to prevent infinite loops.
```

.
Here is the screenshot
<img width="776" height="461" alt="Image" src="https://github.com/user-attachments/assets/d21ac32b-065c-4d93-820f-bf64b245e438" />

### Steps to Reproduce

Create the long chat with streaming.

### Expected Behavior

The streamdown compoenent should not generate any error while streaming.

### Actual Behavior

When I stream the ai response, `streamdown` component generate above specified error while `reat-markdown` is not generating anykind of error.

### Code Sample

```tsx

```

### Streamdown Version

1.2.0

### React Version

19.0.0

### Operating System

None

Nishchit14:
The app works smoothly with `react-markdown`, but when I switch to `streadown`’s `Response` component, after streaming for about 15 seconds, it starts showing the `Maximum update depth exceeded` error.


```tsx
<MessageContent className='text-[0.938rem] group-[.is-assistant]:bg-background'>
        {message.parts.map((part, i) => {
          switch (part.type) {
            case 'text':
              return (
                <div key={`${message.id}-${i}`}>

                      {/* working fine with react-markdown*/}
                      <Markdown content={part.text}
                            className={cn(
                              'prose prose-base dark:prose-invert relative min-w-full bg-transparent p-0',
                              'selection:bg-blue-100 selection:text-blue-900'
                            )} />

                      {/* this is throwing error*/}
                      {/* <Response>{part.text}</Response> */}
                </div>
              );  
```

Nishchit14:
I think the issue is more specific to the ai-sdk. This solution would work in this case https://ai-sdk.dev/docs/troubleshooting/react-maximum-update-depth-exceeded

I will observe this  for the next couple of days in production. Once I feel confident, I will close this issue.

mvberg:
> I will observe this for the next couple of days in production. Once I feel confident, I will close this issue.

@Nishchit14 how are things looking for you after setting `experimental_throttle`?

Nishchit14:
@mvberg It is performing well. I think the issue is resolved now.

PR #530, opened 2026-06-02
fix(streamdown): use useDeferredValue for blocks-state, prevent React #185 cascade

cristicretu:
We hit React #185 with Streamdown in production at [Anara](https://anara.com). About 15 events/hr at peak across thousands of users. Bumping `experimental_throttle` to 100ms (per #140) closed roughly 86% of those, but the rest needed a code change.

The cascade sits in `index.tsx`:

```tsx
const [displayBlocks, setDisplayBlocks] = useState<string[]>(blocks);
useEffect(() => {
  if (mode === "streaming" && !animatePlugin) {
    startTransition(() => setDisplayBlocks(blocks));
  } else {
    setDisplayBlocks(blocks);
  }
}, [blocks, mode]);
```

Each streaming token gives `blocks` a new ref, the effect fires, `setDisplayBlocks` queues an update. `startTransition` lowers the priority but doesn't drop intermediates, so under SSE bursts you stack 50+ updates inside one commit and trip React's nested-update guard.

`useDeferredValue` does the same job without the setStates:

```tsx
const deferredBlocks = useDeferredValue(blocks);
const blocksToRender =
  mode === "streaming" && !animatePlugin ? deferredBlocks : blocks;
```

First render returns `blocks` directly so SSR hydration is unchanged. The `animatePlugin` path stays synchronous because the plugin needs the freshest content per render.


marcoaleixoo:
Hey! Still seeing a lot of this in production as well

farnabaz:
@marcoaleixoo  did you face same issue with Streamdown 2.6?

@cristicretu I would also appreciate if you could check 2.6 and see if you face it or not

cristicretu:
@farnabaz couldn't repro on 2.6! feel free to close

farnabaz:
@cristicretu Thanks for checking.
Then, I close this one, feel free to reopen or drop another PR if you face it in future.

PR #493, opened 2026-04-02
fix(animate): serialize stagger delays across sibling blocks to prevent concurrent animation

sleitor:
## Summary

Fixes #482

When streaming markdown with multiple blocks, animations across sibling blocks revealed concurrently — a list could still be animating while the next paragraph was already fading in.

## Root Cause

All blocks shared a single `AnimatePlugin` instance with a fixed `startIndex=0`. When a new block appeared, its first word started animating at delay=0ms, regardless of what was animating in the previous block.

Additionally, `lastRenderNewWordCount` was never written back to `renderState` in the rehype closure.

## Fix

Introduces `AnimateCursor` — a lightweight shared counter (`{ current: number }`) that:

1. Resets to `0` at the start of each React render pass (in `Streamdown`)
2. Each block's `AnimatePlugin` reads `cursor.current` as its `startIndex` before animating
3. After animating, the plugin increments `cursor.current` by the number of newly animated words

This chains stagger delays automatically across all sibling blocks in render order without any manual `setStartIndex` wiring. Each block gets its own `AnimatePlugin` instance so `prevContentLength` tracking remains per-block independent.

## Changes

- `animate.ts`: Add `AnimateCursor` type + `createAnimateCursor()`, add `cursor` option to `createAnimatePlugin()`, fix missing `lastRenderNewWordCount` write
- `index.tsx`: Replace single shared plugin with per-block plugins array + shared cursor; extract `getBlockPlugins()` helper to stay within biome complexity limits
- `animate.test.ts`: Add 4 cursor chaining tests (36 tests total, all passing)

PR #531, opened 2026-06-03
fix(animate): rework streaming animation into a staged reveal

### LINKED ISSUE #482: Streaming animation reveals multiple markdown sections concurrently instead of serializing section animation
### Bug Description

Streamed markdown can animate multiple sections at once.

For example, while a list is still animating in, the next paragraph or heading can already start animating too. The content order is correct, but the visual reveal feels wrong because multiple sections are “live” at the same time.

Image

### Steps to Reproduce

1. Stream structured markdown with multiple sections, for example:
 - a paragraph
 - a short intro line
 - a bullet list
 - a closing paragraph
2. Watch the response while it is still streaming.
3. Notice that the next paragraph or heading begins animating before the earlier list or section has visually finished.

### Expected Behavior

The list finishes animating before the next paragraph

### Actual Behavior

Image

### Code Sample

```tsx

```

### Streamdown Version

2.5.0

### React Version

19.2.2

### Node.js Version

_No response_

### Browser(s)

_No response_

### Operating System

None

### Additional Context

_No response_

isaacwasserman:
## Description

**Problem 1** (https://github.com/vercel/streamdown/issues/482) — sections animated on top of each other. While a list was still
fading in, the next paragraph or heading would already start animating, so
several parts of the message were “alive” at once and the reveal felt chaotic.

**Problem 2** — rich content animated out of order, or not at all. With
formatting like bold, italics, or links, often only the last word animated while
everything before it popped in instantly, so the reveal looked random.

**Problem 3** — Despite animating in word-by-word, the layout (height of the markdown component) adjusts immediately, causing large amounts of chunkily growing foot space.

How it’s fixed:
- The animation now treats the whole message as one ordered sequence and reveals it a chunk at a time. Each word fades in on its turn, top to bottom, regardless of formatting. In other words, section `i` always animates before section `i+1`, no matter when they were added to the queue.
- Additionally, rather than purely animating the opacity of text, by default we now animate the `display` CSS property to prevent hidden segments from impacting layout.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Related Issues

Fixes #482
Related to #493

## Changes Made

- Streaming content now reveals as one global, ordered sequence — section `i` always animates before section `i+1`, so sibling sections no longer animate at the same time.
- Every word animates in order regardless of formatting (bold, italics, links); code blocks, images, and math reveal as a single unit.
- Unrevealed segments collapse via `display` instead of only fading opacity, so the component grows in step with the reveal instead of reserving the full height up front.
- Already-revealed content reconciles in place — no remounting or re-animation as more text streams in.
- Added an opt-in `animated={{ reserveSpace: true }}` to keep the previous opacity-only behavior (reserve layout for unrevealed content).
- Updated the animation docs. Public API (the `animated` prop, `createAnimatePlugin`, CSS classes/keyframes) is unchanged.

## Testing

- [x] All existing tests pass

### Test Coverage

974 tests passing (77 files). New suites cover the reveal transform, the scheduler, an end-to-end streamed reveal, StrictMode/concurrent idempotency, and the `reserveSpace` toggle; the list re-animation regression test was rewritten for the new model.

## Checklist

- [x] I have made corresponding changes to the documentation

## Changeset

## Additional Notes

- No public API changes. The behavior change is non-breaking: when `animated` is enabled, content is wrapped in lightweight keyed spans that persist for stable reconciliation, and unrevealed text is collapsed with `display: none` by default (use `reserveSpace: true` for the previous opacity-only layout). Inline `code` now animates with the surrounding text.
- This is an alternative approach to #493: instead of per-block plugins chained by a shared cursor, it keeps one shared plugin and orders all blocks through a single controller over a global segment sequence.


farnabaz:
Thanks for the PR, I'm closing this in favor of https://github.com/vercel/streamdown/pull/581 which used the idea of this PR

PR #533, opened 2026-06-03
fix: prevent code block line collapse when lineNumbers is disabled

aradhyacp:
## Description
Fixes an issue where code block lines would collapse into a single visual line when `lineNumbers={false}`.

Ensures each line is always rendered as a block element, preserving line breaks regardless of line number settings.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Related Issues

Fixes #532
Closes #532
Related to #532

## Changes Made

- Ensured each code line `<span>` always has `display: block`
- Separated base line styling from line-number-specific styling
- Prevented line-number classes from being the only source of block display
- Fixed incorrect Tailwind class syntax where background classes were missing closing `)`:
  - `bg-[var(--sdm-bg,inherit]` → `bg-[var(--sdm-bg,inherit)]`
  - `dark:bg-[var(--shiki-dark-bg,var(--sdm-bg,inherit)]` → `dark:bg-[var(--shiki-dark-bg,var(--sdm-bg,inherit))]`

## Testing

- [x] All existing tests pass
- [x] Added new tests for the changes
- [x] Manually tested the changes

### Test Coverage

Manually tested the Code rendered

How I tested it locally:

I wanted to verify my changes, so I linked the local package into a test React app:
```
# In the streamdown repo
pnpm build
pnpm link

# In my test project
pnpm link streamdown
```

Then I used the following test component:

```tsx

import { Streamdown } from "streamdown";

import { code } from "@streamdown/code";

const App = () => {
  const markdown = `
# Project Setup

\`\`\`bash
cd ~/code/project
npm install
npm run dev

# Build
npm run build

# Test
npm test
npm run test:watch

# Lint
npm run lint
npm run lint:fixq
\`\`\`

## Notes

- Ensure Node.js is installed.
- Verify environment variables before starting.
- Run tests before deployment.
- Monitor logs in production.
`;

  return (
    <div className="App">
      <div className="mx-auto mt-5 w-200">
        <Streamdown plugins={{code}} lineNumbers={false}>{markdown}</Streamdown>
      </div>
    </div>
  );
};

export default App;
```

### Before:
image

### After:
image

## Checklist

- [x] My code follows the project's code style
- [x] I have performed a self-review of my own code
[...]

PR #534, opened 2026-06-08
fix(code-block): apply block display to line spans when lineNumbers={false}

### LINKED ISSUE #532: Code block lines collapse when lineNumbers={false}
### Bug Description

When rendering a fenced code block with multiple lines and `lineNumbers={false}`, Streamdown displays the code as a single visual line instead of preserving line breaks.
 
Repro:
 
```bash
cd ~/code/project
npm run dev
```
 
Expected output: two separate lines.
 
Actual output: the two lines render adjacent to each other, like:
 
```bash
cd ~/code/projectnpm run dev
```
 
Observed with `streamdown@2.5.0` and `@streamdown/code@1.1.1`.
  
From inspecting the rendered HTML, Streamdown emits one wrapper `<span>` per highlighted source line, but when `lineNumbers={false}` those line spans are inline and no newline text nodes are inserted between them. With `lineNumbers={true}`, the line spans receive block-level styling, so the issue does not appear.
  
Likely fix: add `block` as a base class for rendered lines. 

### Steps to Reproduce

Repro:

Set `lineNumbers={false}`

```bash
cd ~/code/project
npm run dev
```

Expected output: two separate lines.

Actual output: the two lines render adjacent to each other, like:

```bash
cd ~/code/projectnpm run dev
```

### Expected Behavior

Set `lineNumbers={false}`
  
```bash
cd ~/code/project
npm run dev
```

### Actual Behavior

Set `lineNumbers={false}`

```bash
cd ~/code/projectnpm run dev
```

### Code Sample

```tsx

```

### Streamdown Version

2.5.0

### React Version

19.2.0

### Node.js Version

22.14.0

### Browser(s)

Chrome

### Operating System

macOS

### Additional Context

The change that removed disabling line numbers (#460) removed the `block` class for each line number. 
  
Exact line where classes are applied [here](https://github.com/vercel/streamdown/blob/main/packages/streamdown/lib/code-block/body.tsx#L116).
  
Line number classes [here](https://github.com/vercel/streamdown/blob/main/packages/streamdown/lib/code-block/body.tsx#L15).
  
This should be as simple as having a `LINE_CLASSES_BASE` and a `LINE_NUMBER_CLASSES_BASE`.

sleitor:
## Problem

When `lineNumbers={false}`, code block line spans had no `className` applied, leaving them with the browser default `display: inline`. This caused all code lines to collapse onto the same visual line instead of each appearing on its own line.

## Root Cause

In `body.tsx`, the `className` prop on each line `<span>` was:
```tsx
className={lineNumbers ? lineNumberClasses : undefined}
```

The `LINE_NUMBER_CLASSES_BASE` includes `'block'` as its first class, but when `lineNumbers` is `false`, that class was never applied.

## Fix

- Add a `LINE_CLASSES_BASE = baseCn('block')` constant for the minimal base class all line spans need
- Add a `lineClasses` memo (prefix-aware, like `lineNumberClasses`)
- Change the ternary to `lineNumbers ? lineNumberClasses : lineClasses` so all spans always get `display: block`

## Test

Added a test case to `code-block-line-numbers.test.tsx` that asserts spans have the `block` class when `lineNumbers={false}`.

Fixes #532

### DIFF
--- a/packages/streamdown/__tests__/code-block-line-numbers.test.tsx
+++ b/packages/streamdown/__tests__/code-block-line-numbers.test.tsx
@@ -31,6 +31,14 @@ describe("line numbers", () => {
       );
     });
 
+    it("applies block display class to spans when lineNumbers={false}", () => {
+      const { container } = render(
+        <CodeBlockBody language="js" lineNumbers={false} result={baseResult} />
+      );
+      const span = container.querySelector("code span");
+      expect(span?.className).toContain("block");
+    });
+
     it("applies counter CSS classes to <code> when lineNumbers={true}", () => {
       const { container } = render(
         <CodeBlockBody language="js" lineNumbers={true} result={baseResult} />
--- a/packages/streamdown/lib/code-block/body.tsx
+++ b/packages/streamdown/lib/code-block/body.tsx
@@ -11,6 +11,9 @@ type CodeBlockBodyProps = ComponentProps<"div"> & {
   lineNumbers?: boolean;
 };
 
+// Base line class string for all line spans (ensures block display)
+const LINE_CLASSES_BASE = baseCn("block");
+
 // Base line numbers class string (merged without prefix for memoization)
 const LINE_NUMBER_CLASSES_BASE = baseCn(
   "block",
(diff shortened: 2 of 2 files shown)


farnabaz:
Closed in favor of https://github.com/vercel/streamdown/pull/533

PR #538, opened 2026-06-12
fix(streamdown): re-render memoized components when node offset changes

julesyoungberg:
## Description

Fixes stale DOM when markdown is updated in the middle of an existing block. Memoized leaf components (`ol`, `ul`, `p`, headings, etc.) use `sameNodePosition` to skip re-renders, but it only compared `line` and `column`. When content changed without updating line/column (only `offset`, e.g. injecting a link mid-line), React kept stale output until Streamdown remounted.

`sameNodePosition` is shared by all memoized markdown components via `sameClassAndNode`, so this fix applies broadly rather than to a single element type.

## Type of Change

## Changes Made

- Added `offset?: number` to `MarkdownPoint` and compare `position.start.offset` / `position.end.offset` in `sameNodePosition` ([`packages/streamdown/lib/components.tsx`](packages/streamdown/lib/components.tsx))
- Added regression test `"should re-render when only offset changes"` ([`packages/streamdown/__tests__/components-rerender.test.tsx`](packages/streamdown/__tests__/components-rerender.test.tsx))
- Added patch changeset ([`.changeset/fair-balloons-share.md`](.changeset/fair-balloons-share.md))

## Testing

- [x] All existing tests pass

```bash
pnpm exec vitest run __tests__/components-rerender.test.tsx
```

All 18 tests in that file pass.

## Checklist

## Additional Notes

No documentation or inline comment changes — this is an internal comparator fix with no API surface change. The new test documents the edge case.

### DIFF
--- a/.changeset/fair-balloons-share.md
+++ b/.changeset/fair-balloons-share.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Compare position.start.offset and position.end.offset in sameNodePosition so mid-block markdown updates aren't skipped by React.memo.
--- a/.changeset/fair-balloons-share.md
+++ b/.changeset/fair-balloons-share.md
@@ -2,4 +2,4 @@
 "streamdown": patch
 ---
 
-Compare position.start.offset and position.end.offset in sameNodePosition so mid-block markdown updates aren't skipped by React.memo.
+Fix stale rendering when in-block markdown content changes during streaming updates.
--- a/packages/streamdown/__tests__/components-rerender.test.tsx
+++ b/packages/streamdown/__tests__/components-rerender.test.tsx
@@ -422,6 +422,39 @@ describe("sameNodePosition edge cases", () => {
 
     expect(container.querySelector("ol")).toBeTruthy();
   });
+
+  it("should re-render when only offset changes", () => {
+    const nodeV1 = {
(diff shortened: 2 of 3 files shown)


dasheidt:
This fixes the case where content grows, but there is a case it cannot reach: a replacement of the **same length** moves no offset either.

```tsx
const { container, rerender } = render(<Streamdown>{"Revenue grew"}</Streamdown>);
rerender(<Streamdown>{"Booking grew"}</Streamdown>);   // same length, same offsets

container.textContent; // "Revenue grew" — with or without this PR
```

`start.offset` is 0 and `end.offset` is 12 before and after, so the comparator still reports "unchanged" and the paragraph keeps rendering the old text. Same for a heading, a list item or a `<strong>` run of unchanged length.

The reason offset does not close it is that any position — line, column or offset — describes where the text came from, not what the component renders. I opened #584, which replaces the position comparison with React's own shallow prop comparison minus `node`, so the same test that motivated this PR passes, plus the same-length case and #570.

Not trying to steamroll this one — if the maintainers prefer the narrower change, the offset fix here is strictly better than what is on `main` today.

farnabaz:
Thanks for PR @julesyoungberg, I'm closing this in favor of https://github.com/vercel/streamdown/pull/584.

PR #526, opened 2026-05-20
fix: serialize Mermaid SVG downloads

omribz156:
## Description

Fixes Mermaid SVG/PNG downloads when Mermaid returns HTML-serialized SVG markup containing tags like `<br>`.

The download path now parses the rendered markup as HTML, selects the SVG node, and serializes it with `XMLSerializer` before saving the SVG or converting it to PNG. That keeps the browser-normalized `<br>` from being written back out as invalid SVG markup.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Related Issues

Fixes #516

## Changes Made

- Added a Mermaid SVG download serializer helper.
- Reused the serialized SVG for both SVG download and PNG conversion.
- Added regression coverage for HTML-style `<br>` markup in Mermaid SVG output.
- Added a patch changeset for `streamdown`.

## Testing

- [x] All existing tests pass
- [x] Added new tests for the changes
- [ ] Manually tested the changes

### Test Coverage

- `corepack pnpm build:packages`
- `corepack pnpm --dir packages/streamdown exec vitest run __tests__/mermaid-utils.test.ts __tests__/mermaid-download.test.tsx`
- `corepack pnpm check packages/streamdown/lib/mermaid/download-button.tsx packages/streamdown/lib/mermaid/utils.ts packages/streamdown/__tests__/mermaid-download.test.tsx packages/streamdown/__tests__/mermaid-utils.test.ts packages/streamdown/__tests__/mermaid.test.tsx .changeset/clean-walls-agree.md`
- `corepack pnpm --dir packages/streamdown test`
- `git diff --check`

## Screenshots/Demos

Not applicable.

## Checklist

- [x] My code follows the project's code style
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings or errors
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have created a changeset (`pnpm changeset`)

## Changeset

- [x] I have created a changeset for these changes

## Additional Notes

Implemented with Codex assistance, with the final patch reviewed and kept focused.

PR #542, opened 2026-06-25
fix(mermaid): inject xmlns:xlink namespace if missing before PNG export

### LINKED ISSUE #541: Mermaid PNG export fails for SVGs with xlink:href because generated SVG is not XML-valid
### Bug Description

Some Mermaid diagrams rendered through Streamdown cannot be exported as PNG. The diagram renders visually in the browser, but clicking the PNG download option does nothing / fails silently.

The issue happens when Mermaid returns SVG that contains attributes such as `xlink:href` without declaring the required `xmlns:xlink` namespace on the root `<svg>` element.

Streamdown’s Mermaid PNG export path converts the returned SVG into an image and then draws it to a canvas. Browsers require the SVG to be valid XML for this image-loading step. If the SVG contains `xlink:href` but no `xmlns:xlink`, the image load fails and PNG export cannot complete.

This is similar to the existing issue class where Mermaid SVG may contain HTML-style `<br>` tags, which are also invalid in XML/SVG serialization unless normalized to `<br />`. https://github.com/vercel/streamdown/issues/516

## Affected packages

Observed with:

```txt
streamdown: 2.5.0
@streamdown/mermaid: 1.0.2
mermaid: 11.x
browser: Chromium / Edge
```

## Minimal Mermaid diagram that reproduces it

````md
```mermaid
C4Context
    title Generic system context
    Person(user, "User", "Uses digital services")
    System(application, "Application", "Web and mobile service platform")
    System_Ext(systemA, "External System A", "Provides account data")
    System_Ext(systemB, "External System B", "Provides customer data")
    Rel(user, application, "Uses")
    Rel(application, systemA, "Reads data from")
    Rel(application, systemB, "Reads data from")
```
````

## Root cause

Mermaid can emit SVG like this:

```xml
<svg id="..." xmlns="http://www.w3.org/2000/svg" ...>
  ...
  <image xlink:href="..." />
  ...
</svg>
```

But the SVG root is missing:

```xml
xmlns:xlink="http://www.w3.org/1999/xlink"
```

Because of that, XML/SVG parsing fails with an error equivalent to:

```txt
The prefix "xlink" for attribute "xlink:href" associated with an element type "image" is not bound.
```

Streamdown currently passes this SVG through unchanged into its PNG export conversion. The browser image load fails, so the canvas conversion never succeeds.

## Expected behavior

Streamdown should normalize Mermaid SVG before display/export, especially before PNG conversion.

If SVG contains `xlink:` and no `xmlns:xlink`, add:

```xml
xmlns:xlink="http://www.w3.org/1999/xlink"
```

## Minimal reproduction

```tsx
import { mermaid } from "@streamdown/mermaid";

const markdown = `
\`\`\`mermaid
C4Context
    title Generic system context
    Person(user, "User", "Uses digital services")
    System(application, "Application", "Web and mobile service platform")
    System_Ext(systemA, "External System A", "Provides account data")
    System_Ext(systemB, "External System B", "Provides customer data")
    Rel(user, application, "Uses")
    Rel(application, systemA, "Reads data from")
    Rel(application, systemB, "Reads data from")
\`\`\`
`;

  return <Streamdown plugins={{ mermaid }}>{markdown}</Streamdown>;
}
```

Steps:

1. Render the markdown above with Streamdown and `@streamdown/mermaid`.
2. Open the Mermaid block download menu.
3. Click `PNG`.

### Expected Behavior

Expected:

- PNG file is downloaded.

### Actual Behavior

Actual:

- PNG export fails / no PNG is downloaded.

### Code Sample

```tsx

```

### Streamdown Version

2.5.0

### React Version

19.2.5

### Node.js Version

24.13.0

### Browser(s)

Chrome

### Operating System

Windows

### Additional Context

@streamdown/mermaid version 1.0.2

sleitor:
## Problem

Mermaid can emit SVG containing xlink-prefixed attributes (e.g. `xlink:href` in C4Context and sequence diagrams) without declaring the `xmlns:xlink` namespace on the root `<svg>` element. The resulting SVG is not valid XML, so browsers refuse to load it as an image, causing PNG export to fail silently.

Minimal reproduction (C4Context diagram):
```mermaid
C4Context
  title Generic system context
  Person(user, "User", "Uses digital services")
  System(application, "Application", "Web and mobile service platform")
```

## Fix

Add `addXlinkNamespaceIfMissing()` in `packages/streamdown/lib/mermaid/utils.ts` which:
1. Checks if the SVG string contains `xlink:` prefixed attributes
2. Checks if `xmlns:xlink` namespace is already declared
3. If the namespace is missing, injects `xmlns:xlink="http://www.w3.org/1999/xlink"` onto the root `<svg>` tag via a lightweight string replacement

`svgToPngBlob()` now normalises the input before base64-encoding it as a data URI, following the same pattern as other Mermaid SVG normalisation fixes (see #516).

## Testing

Added 6 new unit tests covering:
- Namespace injection when missing
- No modification when namespace already present
- No modification when no xlink: attributes
- Only root `<svg>` tag is modified (not nested)
- Various xlink attribute types (xlink:href, xlink:actuate, xlink:title)
- Integration test verifying svgToPngBlob encodes the fixed SVG

Fixes #541

### DIFF
--- a/packages/streamdown/__tests__/mermaid-utils.test.ts
+++ b/packages/streamdown/__tests__/mermaid-utils.test.ts
@@ -1,7 +1,11 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { svgToPngBlob } from "../lib/mermaid/utils";
+import { addXlinkNamespaceIfMissing, svgToPngBlob } from "../lib/mermaid/utils";
 
 const BASE64_SVG_DATA_URL_REGEX = /^data:image\/svg\+xml;base64,/;
+const XLINK_NS_ATTR_RE =
+  /<svg[^>]+xmlns:xlink="http:\/\/www\.w3\.org\/1999\/xlink"/;
+const XLINK_NS_GLOBAL_RE = /xmlns:xlink="http:\/\/www\.w3\.org\/1999\/xlink"/g;
+const SVG_OPEN_ATTR_RE = /<svg[^>]*/;
 
 describe("svgToPngBlob", () => {
   let mockCanvas: any;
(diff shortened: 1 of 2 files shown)


farnabaz:
Thanks for PR @sleitor,
#526 already fixed this. It runs Mermaid SVG/PNG downloads through `serializeSvgForDownload()`, which re-parses the SVG as HTML and re-serializes it with `XMLSerializer`. That binds missing `xlink` namespaces (e.g. `xlink:href` without root `xmlns:xlink`) so the SVG is valid XML and PNG export works.

closing this in favor of #526

PR #544, opened 2026-06-25
feat: support separate mermaid export config

jaytcarlson:
## Summary
- add a new `mermaid.exportConfig` option so Mermaid SVG/PNG downloads can use a different config from on-screen rendering
- wire Mermaid download controls to use `exportConfig ?? config` for image exports while preserving existing `.mmd` source export behavior
- add tests and docs covering the new option, plus a changeset for the `streamdown` package

## Why
Apps often need dark-mode Mermaid diagrams in UI but light-theme files for sharing/export. This change supports that directly without custom renderers or DOM workarounds.

## Test plan
- [x] `corepack pnpm --filter streamdown test -- __tests__/mermaid-download.test.tsx`
- [x] `corepack pnpm check`
- [x] `corepack pnpm --filter streamdown test -- __tests__/mermaid-download.test.tsx`

### DIFF
--- a/.changeset/short-points-marry.md
+++ b/.changeset/short-points-marry.md
@@ -1,0 +1,7 @@
+---
+"streamdown": minor
+---
+
+Add `mermaid.exportConfig` to support separate Mermaid configuration for SVG/PNG downloads.
+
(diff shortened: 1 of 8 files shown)


jaytcarlson:
Closing for now because of Mermaid singleton/global-config constraints; this needs an isolated export path (worker/server). 👋

PR #546, opened 2026-06-27
fix(streamdown): disable singleTilde in remark-gfm to prevent ~ from rendering as strikethrough

### LINKED ISSUE #545: Single tilde ~ incorrectly rendered as strikethrough (no way to configure singleTilde: false)
## Problem

Streamdown 2.5.0 internally uses `remark-gfm` with a default empty config `{}`. This means `remark-gfm`'s strikethrough extension runs with `singleTilde: true` by default, causing a single `~` character to be parsed as strikethrough instead of being rendered as a plain character.

In AI chat scenarios, model output frequently contains `~` as a range separator or path component (e.g. `25~30`, `~500ms`, `C:\Users\~\Documents`). These get incorrectly rendered as striethrough.

## Reproduction

```tsx
import { Streamdown } from streamdown;

<Streamdown mode=static>25~30, 31~35</Streamdown>
```

**Expected:** `25~30, 31~35` (plain text)

**Actual:** `25<del>30, 31</del>35` (strikethrough)

## Cause

In `node_modules/streamdown/dist/chunk-BO2N2NFS.js`, the default remark plugins are:

```js
var Ks = { gfm: [remarkGfm, {}], codeMeta: un };
```

`remark-gfm`'s strikethrough extension defaults to `singleTilde: true`, which allows `~text~` to trigger strikethrough. Additionally, `@streamdown/cjk` further relaxes CJK boundary matching for strikethrough, making it even easier to accidentally trigger in Chinese text.

## Expected Behavior

Streamdown should either:

1. Provide a way to configure `remark-gfm` options (e.g. `singleTilde: false`), or
2. Default `singleTilde` to `false`, consistent with GitHub's own behavior: GitHub only renders `~~text~~` as strikethrough; a single `~` is treated as a plain character.

## Current Workaround

Currently, consumers cannot override this because Streamdown's `remarkPlugins` prop appends to the internal plugin list rather than replacing it. The internal `remark-gfm` instance with `singleTilde: true` always runs first.

sleitor:
## Summary

Fixes #545

Single tilde (`~foo~`) was incorrectly rendered as strikethrough because `remark-gfm` defaults to `singleTilde: true`. This PR sets `singleTilde: false` in the default `remark-gfm` configuration, consistent with GitHub GFM behavior where only double tildes (`~~text~~`) create strikethrough.

## Changes

- **`packages/streamdown/index.tsx`**: Changed default `remarkGfm` options from `{}` to `{ singleTilde: false }`
- **`packages/streamdown/__tests__/single-tilde.test.tsx`**: Added tests verifying single `~` is not rendered as strikethrough while `~~` still works
- **`.changeset/fix-single-tilde-545.md`**: Changeset for patch release

## Testing

- All 985 existing tests pass
- 3 new tests added specifically for single tilde behavior

### DIFF
--- a/.changeset/fix-single-tilde-545.md
+++ b/.changeset/fix-single-tilde-545.md
@@ -1,0 +1,12 @@
+---
+"streamdown": patch
+---
+
+fix: disable singleTilde in remark-gfm to prevent single ~ from rendering as strikethrough
+
+Single tilde (`~foo~`) should not be interpreted as strikethrough markup.
+This aligns with GitHub GFM behavior where only double tildes (`~~text~~`)
+create strikethrough. Set `singleTilde: false` in the default remark-gfm
+configuration.
+
+Fixes #545
--- a/packages/streamdown/__tests__/single-tilde.test.tsx
+++ b/packages/streamdown/__tests__/single-tilde.test.tsx
@@ -1,0 +1,39 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { Streamdown } from "../index";
+
+describe("Single tilde handling (#545)", () => {
+  it("should not render single tilde as strikethrough", () => {
+    const content = "~foo~ is not strikethrough";
+    const { container } = render(<Streamdown>{content}</Streamdown>);
+
+    // Single tilde should NOT create a <del> element
+    const del = container.querySelector("del");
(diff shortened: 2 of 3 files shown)


farnabaz:
closing in favor of https://github.com/vercel/streamdown/issues/545#issuecomment-5269560643

Issue #543, opened 2026-06-25
Make tailwind optional


mindplay-dk:
### Problem Statement

I don't use tailwind, and I don't want to introduce tailwind into a project that uses a completely different design system, just for one component.

I've seen #13 and I am already using the package without tailwind.

But why does it need to be tailwind by default?

It doesn't look like there is any real dependency on tailwind at all.

### Proposed Solution

If you would drop the tailwind requirement and make it optional, we could have a smaller bundle size.

The default components for all standard HTML elements can literally just be:

```js
({ node, children }) =>
  createElement(node.tagName, node.properties, children)
```

This would be totally fine for most projects - many sites will already have a default typography, and this will "just work", and will look like the rest of the content on the web site.

Ideally, I would suggest you should export two components, e.g. `Streamdown` for tailwind users (to retain backwards compatibility) and `StreamdownUnstyled` for the rest of us.

With a bit of internal refactoring, `Streamdown` can essentially be a wrapper around `StreamdownUnstyled`, such that tree-shaking can shed the tailwind overhead from the bundle.

With an internal factory function for the component, you could easily omit `tailwind-merge` from the unstyled component and shed this from the bundle as well.

As an side, the `className` property doesn't work very well, probably not even for tailwind users - setting it to `hello` produces a `class` attribute like:

    space-y-4 whitespace-normal [&>*:first-child]:mt-0 [&>*:last-child]:mb-0 hello

Even tailwind users will need to undo this styling by overriding each of the properties - and the rest of us just have these unused classnames in our resulting HTML. I think it would be simpler for everyone if `className` would *replace* rather than add to the attribute. Anyhow, that's just my opinion - if there was an unstyled version of the component, this wouldn't matter to me.

Lastly, I would suggest adding a `defaultComponent` property - since we have the `allowedTags` property, which can allow tags that don't have a defined render-function in `components`, this would act as a fallback for those. It would also mean that the default configuration for the unstyled component can literally just be the function I showed above, with nothing configured for `components` by default at all.

### Alternatives Considered

This is the obvious answer:

```
              <Streamdown
                key={index}
                className="markdown-content"
                components={{
                  h1: ({ children }) => <h1>{children}</h1>,
                  h2: ({ children }) => <h2>{children}</h2>,
                  h3: ({ node, children }) =><h3>{children}</h3>,
                  h4: ({ children }) => <h4>{children}</h4>,
                  h5: ({ children }) => <h5>{children}</h5>,
[...]

PR #548, opened 2026-07-03
feat: add `fallbackComponent` fallback prop

sleitor:
## Summary

Adds a new `defaultComponent` prop to `<Streamdown>` — an additive, backwards-compatible fallback renderer for any HTML tag or allowed custom tag that does not have an explicit entry in the `components` map.

### What it does

When `defaultComponent` is set, it is used for:

- Custom tags declared via `allowedTags` that have no matching key in `components`
- Standard HTML tags not covered by the built-in Tailwind component set (e.g. `<span>`, `<div>`, `<section>`)

Explicit `components` entries always take precedence over `defaultComponent`.

### Why

This directly addresses the boilerplate-reduction use case from Refs #543: users who want a pass-through/unstyled renderer no longer need to enumerate every tag individually.

```tsx
// Before: 25+ explicit component entries just to get plain HTML
<Streamdown components={{ h1: ..., h2: ..., p: ..., ... }}>{md}</Streamdown>

// After: one prop, covers everything not explicitly overridden
<Streamdown
  defaultComponent={({ node, children, ...props }) =>
    createElement(node!.tagName, props, children)
  }
>
  {md}
</Streamdown>
```

### Scope note

This PR implements **only the `defaultComponent` fallback** — the clean, additive, backwards-compatible piece of Refs #543. The broader `Streamdown` / `StreamdownUnstyled` split and `className`-merge semantic change described in the issue are architecture-level decisions that involve breaking changes and were intentionally left out of this PR.

### Implementation details

- The `Proxy` approach is used on the components map so that `hast-util-to-jsx-runtime`'s `hasOwnProperty` checks (via `own.call`) see the fallback for uncovered tag names — both `getOwnPropertyDescriptor` and `get` traps are intercepted.
- `allowedTags` entries are also eagerly registered in the map for clarity and correct Block-comparator invalidation.
- When `defaultComponent` is `undefined`, behaviour is 100% unchanged (no Proxy created).

### Checklist

- [x] Backwards compatible — no behaviour change when prop is absent
- [x] Tests added (`__tests__/default-component.test.tsx`, 9 tests)
- [x] All 991 existing tests pass
- [x] Biome lint + format clean (`pnpm check`)
- [x] TypeScript clean for changed files
- [x] Changeset added (minor bump)
- [x] README updated with usage example

Refs #543

### DIFF
--- a/.changeset/feat-default-component-fallback.md
+++ b/.changeset/feat-default-component-fallback.md
@@ -1,0 +1,26 @@
+---
+"streamdown": minor
+---
+
+feat: add `defaultComponent` fallback prop for unstyled/custom tag rendering
+
+Adds a new `defaultComponent` prop to `<Streamdown>`. When provided, it is
+used as a fallback renderer for any HTML tag or allowed custom tag that does
+not have an explicit entry in the `components` map. This enables unstyled /
+passthrough rendering without enumerating every tag:
+
+```tsx
+<Streamdown
+  defaultComponent={({ node, children, ...props }) =>
+    createElement(node!.tagName, props, children)
+  }
+>
+  {markdown}
+</Streamdown>
+```
+
+`defaultComponent` applies to custom tags declared via `allowedTags` (no
+explicit component entry) and to standard HTML tags absent from the built-in
+Tailwind component set. Explicit `components` entries always win.
+
+Refs #543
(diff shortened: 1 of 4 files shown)


farnabaz (approved):
Thanks

PR #470, opened 2026-03-17
fix(table): add table-wrapper to fullscreen portal so copy/download work

sleitor:
## Summary

Fixes copy and download buttons being unresponsive when a table is in fullscreen mode.

## Root Cause

`TableCopyDropdown` and `TableDownloadDropdown` both use:

```ts
dropdownRef.current?.closest('[data-streamdown="table-wrapper"]')
```

to locate the `<table>` element. When the fullscreen overlay is rendered via `createPortal` (appended to `document.body`), the DOM ancestry is severed — none of the portal's ancestors have `data-streamdown="table-wrapper"`, so `closest()` returns `null` and the copy/download actions silently fail.

## Fix

Add `data-streamdown="table-wrapper"` to the inner `role="presentation"` div inside the fullscreen portal. This restores the ancestor chain that the dropdowns rely on to find the table.

## Testing

Added a test to `table-fullscreen.test.tsx` verifying that `[data-streamdown="table-wrapper"]` and `[data-streamdown="table"]` are accessible inside the fullscreen overlay. All 983 tests pass.

Closes #468

Issue #468, opened 2026-03-17
It cannot be copied or downloaded in full-screen table.


nanwy:
### Bug Description

<img width="157" height="194" alt="Image" src="https://github.com/user-attachments/assets/91ceca1d-fdb1-472f-89e8-b7d1878b55c2" /> Clicking the button is not responding.

### Steps to Reproduce

After rendering the table in markdown format, click the Full Screen button and click the Copy or Download button in full screen.

### Expected Behavior

copied or downloaded

### Actual Behavior

Clicking the button is not responding in full-screen table.

### Code Sample

```tsx

```

### Streamdown Version

2.4.0

### React Version

19

### Node.js Version

24

### Operating System

None

minzifang:
I can also reproduce this in Dify's chat table fullscreen mode: the download dropdown opens, but clicking CSV or Markdown does not trigger a download; non-fullscreen table download works.

PR #552, opened 2026-07-13
fix(mermaid): make controls reachable in the fullscreen view

dkachur1:
## Summary

When a Mermaid diagram is expanded to fullscreen, there is no way to download it — the download button disappears entirely.

## Root cause

`MermaidFullscreenButton` renders its overlay via `createPortal(..., document.body)`. The block's action toolbar (`MermaidDownloadDropdown`, copy button, `MermaidFullscreenButton` itself) is rendered as a sibling of the diagram inside `components.tsx`, *outside* that portal. Since the portal is `fixed inset-0 z-50`, it visually covers that toolbar, and structurally the toolbar was never included in the portal's own markup — so once fullscreen is open there's no download affordance reachable at all.

This is the same class of bug already fixed for tables in #470 (`fix(table): add table-wrapper to fullscreen portal so copy/download work`), closed by #468 — except for tables the buttons were present in the portal but a `closest()` DOM lookup failed; for Mermaid the fullscreen view never rendered a download control in the first place.

## Fix

Render `MermaidDownloadDropdown` inside the fullscreen portal, next to the exit button, respecting the existing `controls.mermaid.download` config (mirrors the same inline boolean-resolution pattern already used for `panZoom` in this file). Propagation is stopped on the wrapping div so clicking download doesn't also trigger the backdrop's close-on-click handler.

## Testing

- Added two tests to `mermaid-fullscreen.test.tsx`:
  - the download button renders inside the fullscreen portal, downloads a PNG on click, and does not close the fullscreen overlay
  - the button is hidden when `controls.mermaid.download` is `false`
- All existing tests in `mermaid.test.tsx`, `mermaid-component.test.tsx`, `mermaid-download.test.tsx`, `mermaid-fullscreen.test.tsx` pass (56/56).
- `biome check` clean on both changed files.

Added a changeset.

### DIFF
--- a/.changeset/mermaid-fullscreen-download.md
+++ b/.changeset/mermaid-fullscreen-download.md
@@ -2,4 +2,4 @@
 "streamdown": patch
 ---
 
-fix(mermaid): render the download control inside the Mermaid fullscreen portal so it's reachable when the diagram is expanded
+fix(mermaid): render the download and copy controls inside the Mermaid fullscreen portal so they're reachable when the diagram is expanded
--- a/.changeset/mermaid-fullscreen-download.md
+++ b/.changeset/mermaid-fullscreen-download.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+fix(mermaid): render the download control inside the Mermaid fullscreen portal so it's reachable when the diagram is expanded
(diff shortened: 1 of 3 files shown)


farnabaz (approved):
Thanks

PR #554, opened 2026-07-14
Preserve line breaks when copying table cells containing <br> elements

aradhyacp:
## Description

Fixes an issue where copying rendered Markdown tables lost line breaks inside cells containing `<br>` elements.

Previously, table extraction relied on `textContent`, which removed `<br>` elements without preserving them as newline characters. This caused multiline cell content to be merged together when copying tables as Markdown, CSV, or TSV.

This change preserves line breaks during table extraction so copied table content maintains the original formatting.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Related Issues

Fixes #553
Closes #553
Related to #553

## Changes Made

- Replaced `textContent` based table cell extraction with a recursive text extraction helper.
- Preserved `<br>` elements by converting them into newline characters during extraction.
- Ensured nested elements containing line breaks are handled correctly.
- Improved copied table output consistency across Markdown, CSV, and TSV formats.
- Added test coverage for single, multiple, and nested `<br>` elements.

## Testing

- [x] All existing tests pass
- [x] Added new tests for the changes
- [x] Manually tested the changes

### Test Coverage

Added tests covering:

- Converting a single `<br>` element into a newline.
- Converting multiple `<br>` elements into newlines.
- Preserving `<br>` elements inside nested elements.

Manually tested the copy functionality using a local React app.

How I tested it locally:

I used the following test component to verify table rendering and copying behavior:

```tsx
import { Streamdown } from "streamdown";
import "streamdown/styles.css";

const App = () => {
  const markdown = `
  | Name | Description |
  | --- | --- |
  | Example | Paragraph one.<br>Paragraph two. |
  | Example | Paragraph one.<br>Paragraph two. |
  | Example | Paragraph one.<br>Paragraph two. |
  | Example | Paragraph one.<br>Paragraph two. |
  | Example | Paragraph one.<br>Paragraph two. |
  `;

  return (
    <div className="App">
      <div className="mx-auto mt-5 w-200">
        <Streamdown>{markdown}</Streamdown>
      </div>
    </div>
  );
};

export default App;
```

## Before Markdown Copy
image

## After Markdown Copy
image

## Before CSV Copy
image

## After CSV Copy
image

## Before TSV Copy

image

## After TSV Copy

image

[...]

PR #555, opened 2026-07-14
fix: preserve <br> line breaks when copying table cells

### LINKED ISSUE #553: Table copy drops line breaks from cells containing <br>
### Bug Description

When copying a rendered Markdown table as Markdown, CSV, or TSV, line breaks inside table cells are lost from the `text/plain` clipboard representation.

The issue appears to be caused by `extractTableDataFromElement()`, which uses `cell.textContent?.trim()`.

For a rendered cell such as `<td>Paragraph one.<br>Paragraph two.</td>`, `textContent` returns `Paragraph one.Paragraph two.`. The `<br>` is therefore lost before the table is serialized.

The HTML clipboard representation still contains the `<br>`, so rich-text paste and plain-text paste produce inconsistent results.

### Steps to Reproduce

1. Render the following Markdown with Streamdown:

   ~~~markdown
   | Name | Description |
   | --- | --- |
   | Example | Paragraph one.<br>Paragraph two. |
   ~~~

2. Open the copy dropdown on the rendered table.
3. Select CSV or TSV.
4. Paste the copied content into a plain-text editor.

The problem can also be observed by reading the `text/plain` clipboard representation, for example with `pbpaste` on macOS.

Minimal example:

~~~tsx
import "streamdown/styles.css";

const markdown = `
| Name | Description |
| --- | --- |
| Example | Paragraph one.<br>Paragraph two. |
`;

  return <Streamdown>{markdown}</Streamdown>;
}
~~~

### Expected Behavior

Line breaks in table cells should be preserved when generating the plain-text Markdown, CSV, or TSV representation.

For CSV, the multiline field should contain a real newline and be quoted:

~~~csv
Example,"Paragraph one.
Paragraph two."
~~~

For TSV or Markdown, the line break should also be represented without merging the two paragraphs into one continuous string.

### Actual Behavior

The line break is removed completely:

~~~text
Example,Paragraph one.Paragraph two.
~~~

The two paragraphs are concatenated without a newline or even a separating space.

Pasting the same clipboard content into an application that reads the `text/html` representation may preserve the line break because the HTML still contains `<br>`. The problem affects the generated `text/plain` representation.

### Code Sample

```tsx

```

### Streamdown Version

2.5.0

### React Version

19.2.7

### Node.js Version

_No response_

### Browser(s)

_No response_

### Operating System

None

### Additional Context

_No response_

AliMahmoudDev:
## Summary

Fixes #553

When copying a rendered Markdown table as Markdown, CSV, or TSV, line breaks inside table cells were lost from the `text/plain` clipboard representation.

### Root Cause

`extractTableDataFromElement()` used `cell.textContent` which collapses `<br>` tags into nothing. For a rendered cell like:

```html
<td>Paragraph one.<br>Paragraph two.</td>
```

`textContent` returns `Paragraph one.Paragraph two.` — the line break is silently dropped.

### Fix

Replace `textContent` with a custom `getCellText()` helper that walks the DOM tree and inserts a newline character (`\n`) for each `<br>` element. This preserves line breaks through the full copy pipeline: extraction → CSV/TSV/Markdown serialization → clipboard.

The existing CSV/TSV/Markdown serializers already handle `\n` characters correctly (CSV wraps in quotes, TSV escapes as `\n`, Markdown passes through), so the fix only needs to happen at the extraction stage.

### Tests Added

- Single `<br>` in a data cell → produces `\n` in extracted text
- Multiple consecutive `<br>` tags → produces `\n\n`
- `<br>` in header cells → produces `\n` in headers

All existing tests continue to pass since `getCellText` produces the same output as `textContent.trim()` for cells without `<br>` tags.

### DIFF
--- a/packages/streamdown/__tests__/table-utils.test.ts
+++ b/packages/streamdown/__tests__/table-utils.test.ts
@@ -119,6 +119,66 @@ describe("Table Utils", () => {
       expect(result.headers).toEqual([]);
       expect(result.rows).toEqual([["Data"]]);
     });
+
+    it("should preserve line breaks from <br> tags in cells", () => {
+      tableElement.innerHTML = `
+        <thead>
+          <tr>
+            <th>Name</th>
+            <th>Description</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>Example</td>
+            <td>Paragraph one.<br>Paragraph two.</td>
+          </tr>
+        </tbody>
+      `;
+
+      const result = extractTableDataFromElement(tableElement);
+
+      expect(result.headers).toEqual(["Name", "Description"]);
+      expect(result.rows).toEqual([["Example", "Paragraph one.\nParagraph two."]]);
+    });
+
+    it("should preserve multiple <br> tags in cells", () => {
+      tableElement.innerHTML = `
+        <thead>
+          <tr>
+            <th>Content</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>Line 1<br><br>Line 3</td>
+          </tr>
+        </tbody>
+      `;
+
+      const result = extractTableDataFromElement(tableElement);
+
+      expect(result.rows).toEqual([["Line 1\n\nLine 3"]]);
+    });
+
+    it("should preserve <br> tags in header cells", () => {
+      tableElement.innerHTML = `
+        <thead>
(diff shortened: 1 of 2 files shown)


farnabaz:
Thanks for the PR, but it already handled in #554.
Closing in favor or https://github.com/vercel/streamdown/pull/554

PR #563, opened 2026-07-22
fix(sanitize): prevent double `user-content-` prefix on footnote ids

helloyou2012:
Footnote and heading ids were rendered with a doubled prefix (e.g. `user-content-user-content-fn-1`) because both `remark-rehype` and `rehype-sanitize` default their `clobberPrefix` to `user-content-`, so
the prefix was applied twice in the pipeline.

remark-rehype emits footnote backref hrefs (`href="#fn-1"`) coordinated with its own `clobberPrefix`, so disabling the prefix upstream would desync hrefs from their targets and break footnote navigation. Disable `clobberPrefix` on rehype-sanitize instead, letting remark-rehype own the single, consistent prefixing of both hrefs and ids.

- index.tsx: set `clobberPrefix: ""` on `defaultSanitizeSchema`
- __tests__/footnotes.test.tsx: add regression test asserting footnote
  ids are prefixed exactly once
- __tests__/allowed-tags.test.tsx: update assertions — user-provided
  ids on custom tags are no longer clobbered by sanitize

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Related Issues

Fixes #
Closes #
Related to #

## Changes Made

- 
- 
- 

## Testing

- [x] All existing tests pass

## Checklist

## Changeset

### DIFF
--- a/.changeset/fix-footnote-double-user-content-prefix.md
+++ b/.changeset/fix-footnote-double-user-content-prefix.md
@@ -1,0 +1,7 @@
+---
+"streamdown": patch
+---
+
+Fix doubled `user-content-` prefix on footnote ids
+
+Footnote list items were rendered with `id="user-content-user-content-fn-1"` because both `remark-rehype` and `rehype-sanitize` default their `clobberPrefix` to `user-content-`, so the prefix was applied twice. Disabled `clobberPrefix` on `rehype-sanitize` so `remark-rehype` remains the single, consistent prefixer of both footnote ids and backref hrefs — restoring working footnote navigation.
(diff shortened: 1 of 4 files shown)


helloyou2012:
Hi @haydenbleasel ! This PR fixes an issue where footnote links can't be navigated to.

  **Problem**: Footnote definition elements have their id prefixed with user-content- twice
  (id="user-content-user-content-fn-1"), while the backref link's href only gets the prefix once (#user-content-fn-1).
  The href and target id no longer match, so clicking a footnote goes nowhere.

  **Root cause**: Both remark-rehype and rehype-sanitize default their clobberPrefix to user-content-, and the two stages
  stack, producing a doubled prefix on ids.

  **Fix**: Disable clobberPrefix on rehype-sanitize so remark-rehype remains the single source of prefixing, keeping ids and
   hrefs in sync.

  Would appreciate a review — thanks!

farnabaz (approved):
LGTM 👍 
Thanks for the PR

PR #573, opened 2026-08-11
fix: remove animation spans after streaming ends

### LINKED ISSUE #570: Animation spans are never removed from settled text — memoized components discard the span-free reparse
## Summary

`streamdown@2.5.0`, React 19.

## Reproduction

```tsx

  );

  rerender(
  );
});
```

## Cause

```ts
// lib/components.tsx
}
```

### Possible fixes

---

```tsx

```

BetterAndBetterII:
## Description

Fix animation spans (`data-sd-animate`) persisting in the DOM after `isAnimating` goes false. Memoized leaf components (`MemoParagraph`, `MemoLi`, etc.) compare only `className` + source position and never compare children, so they discard the span-free reparse that Block produces when the animate rehype plugin is removed from the pipeline.

## Root cause

`Block` comparator correctly detects the rehype pipeline change and re-parses without the animate plugin. But the resulting span-free HAST never commits because leaf memo comparators (`sameClassAndNode`) see identical class + position and return "equal" — React throws away the new children.

## Fix

Include `isAnimating` in block keys. When streaming ends, keys change and React mounts fresh subtrees, allowing the span-free reparse to commit. During streaming (`isAnimating` stays true), keys are stable — no hot-path penalty.

## Type of Change

## Related Issues

Fixes #570

## Changes Made

- `packages/streamdown/index.tsx`: Include `isAnimating` in block key computation so React remounts fresh subtrees when the animate rehype plugin is removed
- `packages/streamdown/__tests__/animate-integration.test.tsx`: New integration test verifying spans are removed when `isAnimating` goes false
- `.changeset/clean-settled-animation-spans.md`: Patch changeset

## Testing

- [x] All existing tests pass (983/983, 74 files)
- [x] Added new integration test: `animate-integration.test.tsx`
- [x] RED confirmed: test fails without fix (19 spans persist)
- [x] GREEN confirmed: test passes with fix (0 spans)
- [x] No regressions in full test suite

## Checklist

- [x] My code follows the project code style

## Additional Notes

The fix only affects the transition when `isAnimating` flips. During active streaming (isAnimating stays true), block keys remain stable — no extra remounting on the hot path. The one-time remount cost when streaming ends is negligible compared to the persistent DOM pollution of stale animation spans.

### DIFF
--- a/.changeset/clean-settled-animation-spans.md
+++ b/.changeset/clean-settled-animation-spans.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Fix animation spans not being removed from settled text when `isAnimating` goes false. Include `isAnimating` in block keys so React mounts fresh subtrees when the animate rehype plugin is removed from the pipeline, allowing the span-free reparse to commit.
--- a/packages/streamdown/__tests__/animate-integration.test.tsx
+++ b/packages/streamdown/__tests__/animate-integration.test.tsx
@@ -1,0 +1,37 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { Streamdown } from "../index";
+
+const FADE = {
+  animation: "fadeIn",
+  duration: 200,
+  easing: "ease-out",
+  sep: "word",
+  stagger: 0,
+} as const;
+
+const MD =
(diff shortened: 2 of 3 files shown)


farnabaz:
Thanks for the PR, however as part of animation refactor in https://github.com/vercel/streamdown/pull/581 this issue fixed.
closing in favor of https://github.com/vercel/streamdown/pull/581

PR #574, opened 2026-08-11
fix: quadratic code-block scan in isInsideCodeBlock

Vikramb1:
## Problem

`isInsideCodeBlock(text, position)` rescans the entire prefix `[0, position)` on every call. Several handlers probe it once per candidate character — the link/image handler calls it for every `[` while walking backwards — so total work is O(brackets × length).

The pathological case is exactly what streaming AI produces: a long, still-unclosed code fence full of brackets, re-repaired on every arriving token. Repro (58k chars):

```ts
const line = "const x = arr[i]; if (map[key]) { list[j] = grid[a][b]; }\n";
remend("\`\`\`ts\n" + line.repeat(1000)); // ~915ms on remend@1.3.0
```

That's ~915ms of main-thread time *per streamed chunk* when rendering a large code block.

## Fix

Build the answer for every position in one linear pass (`buildCodeBlockLookup`, mirroring the original scan's control flow including escaped-backtick and triple-backtick handling), and cache it per text with a single-entry cache — handlers repeatedly probe the same string within one `remend()` call, so each probe becomes O(1) after one O(n) build. No signature changes; all six consuming handlers benefit.

Same repro after the fix: **~0.4ms** (benchmark added, `iterations: 10` to keep CI fast).

## Behavior parity

- New `__tests__/code-block-utils.test.ts` includes a verbatim copy of the previous implementation as a reference oracle and asserts equality at **every position** across mixed inputs (fences, inline code, escaped backticks, unclosed spans), plus a cache-correctness test with alternating texts.
- Positions past `text.length` resolve to the end-of-scan state, matching the old loop's behavior.
- Full existing suites pass unmodified: remend 464/464, streamdown 982/982.

## Notes

- Changeset included (patch).
- `pnpm check` / `check-types` clean. `pnpm build` succeeds for all library packages (`apps/test` prerender needs AI Gateway credentials I don't have as an external contributor).

### DIFF
--- a/.changeset/quadratic-code-block-scan.md
+++ b/.changeset/quadratic-code-block-scan.md
@@ -1,0 +1,5 @@
+---
+"remend": patch
+---
+
+Fix quadratic code-block scanning. `isInsideCodeBlock` now builds a linear-time position lookup (cached per text) instead of rescanning the whole prefix on every call, so handlers that probe many positions no longer degrade quadratically. Repairing an unclosed, bracket-heavy 58k-character code block drops from ~915ms to ~0.4ms per call.
--- a/packages/remend/__benchmarks__/remend.bench.ts
+++ b/packages/remend/__benchmarks__/remend.bench.ts
@@ -279,3 +279,20 @@ ${"Regular paragraph text with some [links](https://example.com) and more conten
     { iterations: 1000 }
   );
 });
+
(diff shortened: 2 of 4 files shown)


farnabaz (approved):
This is a big improvement 👍 
Thanks

PR #576, opened 2026-08-12
fix(remend): drop lookbehind from single-tilde pattern for Safari < 16.3

### LINKED ISSUE #519: Streamdown crashes on iOS 16 / Safari < 16.3 due to lookbehind regex in `remend` and `mdast-util-gfm-autolink-literal`
## Summary

Streamdown fails to initialize on **iOS 16.0–16.2** (and any Safari/WebKit < 16.3) because two regexes in the dependency chain use lookbehind assertions `(?<=...)` / `(?<!...)`, which JSCore on those versions does not support. The error is thrown at **module top-level evaluation**, with no `try/catch`, so the entire app chunk fails to export and the page goes blank.

## Reproduction

Render any markdown with `<Streamdown>` in a Next.js (or any bundler that ships the package as ESM/CJS into the browser) and load it on an iOS 16.0–16.2 device or Safari Technology Preview pinned to that JSC version.

```tsx

export default function Page() {
  return <Streamdown>hello</Streamdown>;
}
```

Console:

```
SyntaxError: Invalid regular expression: invalid group specifier name
  RegExp
  (anonymous) — _app-<hash>.js:173:1032
  __webpack_require__ — webpack-<hash>.js:1:148
  ...
Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'e.getInitialProps')
```

The second error is downstream: the chunk's module factory threw before assigning exports, so Next.js sees `e === undefined` when reading `getInitialProps`.

## Affected versions

- `streamdown@2.5.0` (latest at time of writing)
- `remend@1.3.0` (bundled via streamdown)
- `mdast-util-gfm-autolink-literal@2.0.1` (pulled in by `streamdown → remark-gfm@4`)

### 1. `remend/dist/index.js` (top-level, no `try/catch`)

```js
var Gn = new RegExp(
  "(?<=[\\p{L}\\p{N}_])~(?!~)(?=[\\p{L}\\p{N}_])",
  "gu"
);
```

Used by the `singleTilde` handler to escape stray `~` characters between word chars. Since the `RegExp` constructor runs at module evaluation, the error throws **before any user code executes** and there is no fallback.

### 2. `mdast-util-gfm-autolink-literal/lib/index.js:135`

```js
[/(?<=^|\s|\p{P}|\p{S})([-.\w+]+)@([-\w]+(?:\.[-\w]+)+)/gu, findEmail]
```

Used inside `transformGfmAutolinkLiterals` to find bare email addresses. The regex literal is compiled when the module loads (or wrapped into `new RegExp(...)` by some bundlers/minifiers — Next.js/SWC in our build ended up calling the constructor, which is why we see the `RegExp` frame in the stack), and again there is no fallback.

> Note: streamdown also bundles `marked@15`, which uses `new RegExp("(?<=1)(?<!1)")` for feature detection — that one **is** wrapped in `try/catch`, so it is not the culprit. It only proves that streamdown's own author is aware lookbehind isn't universally supported.

## Why `react-markdown` users don't hit this

`react-markdown@8` resolves to `remark-gfm@3` → `mdast-util-gfm-autolink-literal@1.x`, which does **not** use lookbehind. `streamdown` jumped to `remark-gfm@4` → `mdast-util-gfm-autolink-literal@2.x` and additionally introduced `remend`, both of which adopted lookbehind. So this regresses any project migrating from `react-markdown` to `streamdown`.

## Suggested fixes

Both lookbehinds can be rewritten without lookbehind while preserving semantics:

### `remend` — change the lookbehind to a consuming capture:

```diff
-var Gn = new RegExp(
-  "(?<=[\\p{L}\\p{N}_])~(?!~)(?=[\\p{L}\\p{N}_])",
-  "gu"
-);
-tn = n => !n || typeof n !== "string" || !n.includes("~")
-  ? n
-  : n.replace(Gn, (r, e) => c(n, e) ? r : "\\~");
+var Gn = new RegExp(
+  "([\\p{L}\\p{N}_])~(?!~)(?=[\\p{L}\\p{N}_])",
+  "gu"
+);
+tn = n => !n || typeof n !== "string" || !n.includes("~")
+  ? n
+  : n.replace(Gn, (r, p, e) => c(n, e + 1) ? r : p + "\\~");
```

The captured left word char is written back; `c()` is offset by 1 so it still points at the `~`. Semantics are identical.

### `mdast-util-gfm-autolink-literal` — drop the lookbehind, rely on the existing `previous()` check:

```diff
-[/(?<=^|\s|\p{P}|\p{S})([-.\w+]+)@([-\w]+(?:\.[-\w]+)+)/gu, findEmail]
+[/([-.\w+]+)@([-\w]+(?:\.[-\w]+)+)/gu, findEmail]
```

`findEmail` already calls `previous(match, true)` which inspects `match.input[match.index - 1]` for `^ | whitespace | punctuation`. `mdast-util-find-and-replace` advances `lastIndex` by 1 on `false` (lib/index.js:158-162), so no infinite loop. Only edge case: lookbehind also accepted `\p{S}` (symbols, e.g. `→`), which `previous()` does not — negligible in practice, and worth the broad-browser support.

If you'd rather not change the regex semantics, **wrap both call sites in `try/catch` and fall back to a non-lookbehind version** so the page at least loads.

## Workaround for affected projects (now)

Until upstream lands a fix, this is patchable in userland with `patch-package`:

```
npm i -D patch-package
# edit node_modules/remend/dist/index.js + node_modules/.../mdast-util-gfm-autolink-literal/lib/index.js with the diffs above
npx patch-package remend
npx patch-package streamdown/mdast-util-gfm-autolink-literal
# add "postinstall": "patch-package" to package.json
```

Verified to fix the white-screen on iOS 16.0/16.1/16.2.

## Environment

- Streamdown: 2.5.0
- Next.js: 15.3.8 (Pages Router, SWC compiler)
- Browsers confirmed broken: Safari 16.0, 16.1, 16.2 (iOS + macOS), WKWebView on iOS 16.0–16.2
- Browsers confirmed working: Safari 16.4+, all Chromium, all Firefox
- Lookbehind support reference: https://caniuse.com/js-regexp-lookbehind (Safari shipped support in 16.4, March 2023)

s-andringa:
Refs #519.

## What

Removes the lookbehind assertion from `remend`'s single-tilde escape pattern:

```diff
-const SINGLE_TILDE_PATTERN = /(?<=[\p{L}\p{N}_])~(?!~)(?=[\p{L}\p{N}_])/gu;
+const SINGLE_TILDE_PATTERN = /([\p{L}\p{N}_])~(?!~)(?=[\p{L}\p{N}_])/gu;
```

The preceding word character is now captured and written back in the replacement, and the offset handed to `isInsideCodeBlock` is shifted by its length so it still points at the `~`. Semantics are unchanged — Unicode support (`\p{L}`, `\p{N}`, so CJK and other scripts) is fully preserved.

## Why

JSCore on iOS 16.0–16.2 / Safari < 16.3 does not support lookbehind and throws `SyntaxError: Invalid regular expression: invalid group specifier name`. Because the pattern is a module-level constant, this throws during module evaluation — before any user code runs — so the whole chunk fails to export and the page goes blank. There is no way for a consumer to catch it. Safari only shipped lookbehind in 16.4 ([caniuse](https://caniuse.com/js-regexp-lookbehind)).

## Scope

This PR deliberately only fixes `remend`. #519 identifies a second lookbehind in `mdast-util-gfm-autolink-literal@2`, pulled in via `remark-gfm@4`. That one is **not** addressed here, because the maintainer has stated the lookbehind is intentional and will not be changed: [syntax-tree/mdast-util-gfm-autolink-literal#10 (comment)](https://github.com/syntax-tree/mdast-util-gfm-autolink-literal/issues/10#issuecomment-3330083240). Projects needing full iOS 16.0–16.2 support will have to patch that dependency themselves (`pnpm patch` / `patch-package`).

So this change does not single-handedly close #519 for every `streamdown` consumer. It does fix it outright for anyone depending on `remend` directly, without dragging in `remark-gfm` — which is my own use case, and the one bit of this that is squarely this repo's own code rather than a third party's.

## Testing

- `pnpm --filter remend test` — 458/458 pass, no changes needed to existing tests (behaviour is identical)
- `pnpm --filter streamdown test` — 991/991 pass
- `pnpm check` — clean
- Built `dist` output verified free of any `(?<` occurrences

Changeset included (`remend`, patch).

### DIFF
--- a/.changeset/fix-lookbehind-safari-16-crash.md
+++ b/.changeset/fix-lookbehind-safari-16-crash.md
@@ -1,0 +1,7 @@
+---
+"remend": patch
+---
+
+Fix crash on iOS 16.0-16.2 / Safari < 16.3 by removing the lookbehind assertion from the single-tilde escape pattern (#519).
+
+JSCore on those versions doesn't support lookbehind (`(?<=...)`) and throws a `SyntaxError` while the module is being evaluated, before any user code runs, so there is no way to catch it. The preceding word character is now captured and written back in the replacement instead.
--- a/packages/remend/__tests__/single-tilde.test.ts
+++ b/packages/remend/__tests__/single-tilde.test.ts
@@ -42,4 +42,18 @@ describe("single tilde escape (#445)", () => {
   it("can be disabled via options", () => {
     expect(remend("20~25°C", { singleTilde: false })).toBe("20~25°C");
   });
+
+  it("should preserve Unicode letters around single ~", () => {
+    expect(remend("日本~語")).toBe("日本\\~語");
+    expect(remend("α~β")).toBe("α\\~β");
+    expect(remend("é~x")).toBe("é\\~x");
+  });
+
+  it("should preserve supplementary-plane Unicode letters around single ~", () => {
+    expect(remend("𐐀~a")).toBe("𐐀\\~a");
+  });
+
+  it("should escape multiple single tildes between letters", () => {
+    expect(remend("foo~bar~baz")).toBe("foo\\~bar\\~baz");
+  });
 });
--- a/packages/remend/src/single-tilde-handler.ts
+++ b/packages/remend/src/single-tilde-handler.ts
@@ -12,7 +12,9 @@ import { isInsideCodeBlock } from "./code-block-utils";
 // - NOT followed by another ~ (to avoid matching ~~)
 // - followed by a word character
 // Uses Unicode-aware \p{L} and \p{N} for CJK and other scripts
-const SINGLE_TILDE_PATTERN = /(?<=[\p{L}\p{N}_])~(?!~)(?=[\p{L}\p{N}_])/gu;
+// Captures the preceding char instead of using a lookbehind: Safari < 16.3
+// throws on lookbehind, which would crash at module evaluation (#519).
+const SINGLE_TILDE_PATTERN = /([\p{L}\p{N}_])~(?!~)(?=[\p{L}\p{N}_])/gu;
 
 export const handleSingleTildeEscape = (text: string): string => {
   if (!text || typeof text !== "string") {
@@ -24,12 +26,14 @@ export const handleSingleTildeEscape = (text: string): string => {
     return text;
   }
 
-  return text.replace(SINGLE_TILDE_PATTERN, (match, offset) => {
+  return text.replace(SINGLE_TILDE_PATTERN, (match, precedingChar, offset) => {
+    const tildeOffset = offset + precedingChar.length;
+
     // Don't escape inside code blocks
-    if (isInsideCodeBlock(text, offset)) {
+    if (isInsideCodeBlock(text, tildeOffset)) {
       return match;
     }
 
-    return "\\~";
+    return `${precedingChar}\\~`;
   });
 };


farnabaz (approved):
Thanks

PR #536, opened 2026-06-08
Fix animate plugin whitespace handling in word splitting (especially inside links)

aradhyacp:
## Description

Fixes an issue in the animate plugin where word-based splitting incorrectly handled whitespace inside inline elements (especially <a> links), causing inconsistent token grouping and unexpected animation spans.

This improves whitespace handling logic so words and spaces are grouped more predictably during transformation, reducing fragmented span output.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Related Issues

Fixes #535
Closes #535
Related to #535

## Changes Made

- Improved word-splitting logic for sep: "word" mode
- Fixed whitespace merging behavior in inline text nodes
- Prevented isolated whitespace from becoming separate animated spans
- Improved consistency of token grouping in nested elements
- Ensured stagger delay calculation remains stable after token merging
- Improved behavior inside inline elements like <a>, <strong>, etc.

## Testing

- [x] All existing tests pass
- [x] Added new tests for the changes
- [x] Manually tested the changes

### Test Coverage

Manually tested the Code rendered

How I tested it locally:

I wanted to verify my changes, so I linked the local package into a test React app:
```
# In the streamdown repo
pnpm build
pnpm link

# In my test project
pnpm link streamdown
```

Then I used the following test component:

```tsx
import { Streamdown } from "streamdown";
import "streamdown/styles.css";


const App = () => {
  const markdown = `Here is a sentence with [a link to example.com](https://example.com) sitting right in the middle, followed by some more trailing words so you can clearly see how the link animates in relative to the text around it. And here's that link again: [a link to example.com](https://example.com)`

  const components = {
  a: ({ children, ...props }: React.ComponentProps<"a">) => (
    <a {...props} style={{ textDecoration: "underline", color: "#2563eb" }}>
      {children}
    </a>
  ),
};
  return (
    <div className="App">
      <div className="mx-auto mt-5 w-200">
        <Streamdown
        components={components}
        isAnimating={true}
        animated={{
          animation: "fadeIn",
          duration: 500,
          stagger: 50,
          easing: "ease-out",
        }}
      >{markdown}</Streamdown>
      </div>
    </div>
  );
};

export default App;
```

### Before:
Present in the issue raised

### After:

[...]

PR #581, opened 2026-08-18
fix: concurrent animation across streaming blocks

### LINKED ISSUE #482: Streaming animation reveals multiple markdown sections concurrently instead of serializing section animation

### LINKED ISSUE #535: Animated text with links that have `text-decoration: underline` show underlines before revealed
### Bug Description

The span chunking that animate does causes links with spaces and `text-decoration: underline` to show the underlines before the word chunks are revealed. This is because the chunking is done like `<span>first</span> <span>word</span>` instead of `<span>first </span><span>word</span>`

### Steps to Reproduce

Reproduction here: https://codesandbox.io/p/devbox/tdnjrs

### Expected Behavior

That the `_` between words doesn't show up until that chunk of the word has animated.

### Actual Behavior

Image

### Code Sample

```tsx

const FULL_TEXT =
  "Here is a sentence with [a link to example.com](https://example.com) " +
  "sitting right in the middle, followed by some more trailing words so " +
  "you can clearly see how the link animates in relative to the text " +
  "around it. And here's that link again: [a link to example.com](https://example.com)";

const components = {
  a: ({ children, ...props }: React.ComponentProps<"a">) => (
    <a {...props} style={{ textDecoration: "underline", color: "#2563eb" }}>
      {children}
    </a>
  ),
};

  return (
    <div>
      <Streamdown
        components={components}
        isAnimating={true}
        animated={{
          animation: "fadeIn",
          duration: 200,
          stagger: 50,
          easing: "ease-out",
        }}
      >
        {FULL_TEXT}
      </Streamdown>
    </div>
  );
}
```

### Streamdown Version

2.5.0

### React Version

19

### Node.js Version

24

### Browser(s)

Chrome

### Operating System

macOS

### Additional Context

_No response_

### LINKED ISSUE #550: Streaming mode without 'animated': block updates deferred via useTransition are starved by sibling urgent updates - markdown freezes at first parse
## Summary

In `streamdown@2.5.0`, a `<Streamdown>` in the default streaming mode **without an `animated` config** pushes its parsed-block state updates through `useTransition`. Those low-priority updates can be **starved indefinitely by sibling urgent updates** — e.g. another component ticking state at animation cadence — so the rendered markdown freezes at its *first* parse and never updates again until the urgent-update pressure stops, even though `children` keeps growing and the component keeps re-rendering.

The practical hit: a chat UI that renders a model's **reasoning/thinking stream** in one plain `<Streamdown>` while the **reply** streams in another (animated, or otherwise driving frequent state updates). The thinking panel freezes at its first delta — often a single character — for the whole turn, then snaps to the full text when streaming ends. It looks exactly like a data bug, and we spent a while exonerating our wire/persistence before finding it here.

## Where in the source

In the main component (paraphrasing the 2.5.0 bundle):

```js
const blocks = useMemo(() => parseMarkdownIntoBlocksFn(processedChildren), [...]);
const [deferredBlocks, setDeferredBlocks] = useState(blocks);
useEffect(() => {
  mode === "streaming" && !animatePlugin
    ? startTransition(() => { setDeferredBlocks(blocks); })   // <-- starvable
    : setDeferredBlocks(blocks);
}, [blocks, mode]);
const rendered = mode === "streaming" ? deferredBlocks : blocks;
```

With `animated` set, updates take the direct `setDeferredBlocks(blocks)` branch and everything streams fine — which is also the workaround.

## Repro sketch

```tsx
function Repro() {
  const [reasoning, setReasoning] = useState("");   // grows ~10x/sec
  const [, tick] = useState(0);                      // sibling urgent updates, 40ms
  useEffect(() => {
    const a = setInterval(() => setReasoning((t) => t + " lorem"), 100);
    const b = setInterval(() => tick((n) => n + 1), 40);
    return () => { clearInterval(a); clearInterval(b); };
  }, []);
  return <Streamdown>{reasoning}</Streamdown>; // freezes at the first parse while `b` runs
}
```

Observed in a real app (Next 16.2.6 / React 19, `streamdown@2.5.0`): instrumented the element with `data-len={text.length}` — the prop grew 22 → 139 chars and the attribute updated live (so the component re-rendered), while the rendered markdown text stayed frozen at 3 chars until the sibling stream ended. Passing `animated={{ animation: "fadeIn", duration: 150, sep: "word" }}` + `isAnimating` fixed it instantly, no other change.

## Expected

Either:
- streaming-mode block updates shouldn't be starvable (e.g. fall back to a direct update if a transition hasn't committed within some budget, or don't route through `useTransition` at all when the update is just "the tail block grew"), or
- the behavior/trade-off (and the `animated` escape hatch) documented — right now un-animated streaming markdown next to any busy sibling silently freezes, which reads as data corruption rather than scheduling.

Happy to provide a full repro repo if useful.

### LINKED ISSUE #570: Animation spans are never removed from settled text — memoized components discard the span-free reparse

farnabaz:
Streaming animation was fighting itself: sibling blocks faded in at the same time, spans stuck around after settle, un-animated streams could freeze next to busy siblings, and underlined links flashed early.

This takes the useful bits from earlier attempts — wall-clock stagger chaining from #493, the “don’t let the queue run away” idea from #531, and whitespace handling from #536 — and keeps the existing model instead of rewriting it.

Stagger now shares one timeline across blocks, compresses under a fast stream instead of stacking forever, tears spans down cleanly when `isAnimating` goes false, and glues trailing spaces into word spans so underlines don’t paint ahead.

No public API changes. `stagger` still works as usual; under heavy load the cascade compresses instead of overlapping or stalling.

Inspired by https://github.com/vercel/streamdown/pull/493, https://github.com/vercel/streamdown/pull/531, and https://github.com/vercel/streamdown/pull/536.

Resolves #482
Resolves #535
Resolves #550
Resolves #570


farnabaz:
@sleitor, @aradhyacp, @isaacwasserman I would appreciate it if you could verify the behavior.

aradhyacp:
@farnabaz Tested it locally and everything looks clean from my side. The behavior looks good with no issues noticed. Thanks :)

PR #583, opened 2026-08-18
fix: update streamed code block content

Kinfe123:
## Summary

- include rendered code content in the Markdown code memo comparison
- reuse the same content extraction for rendering and memoization
- add a regression test for replacing a streamed fenced block in place

## Why

`MemoCode` previously compared only `className` and source position. Replacing a streamed code block without moving it in the document, such as changing `first` to `second`, kept the same source position and caused React to retain the stale rendered code. Comparing the extracted code text preserves the existing optimization for unchanged blocks while allowing changed blocks to update.

## Validation

- `pnpm build:packages`
- `pnpm --filter streamdown test` (73 files, 1,019 tests)
- `pnpm check`
- `pnpm --filter streamdown build`

### DIFF
--- a/.changeset/fresh-code-blocks-update.md
+++ b/.changeset/fresh-code-blocks-update.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Fix streamed code blocks remaining stale when their content changes without moving in the document.
--- a/packages/streamdown/__tests__/code-block-memo.test.tsx
+++ b/packages/streamdown/__tests__/code-block-memo.test.tsx
@@ -18,6 +18,30 @@ vi.mock("../lib/code-block", () => ({
 }));
 
 describe("MemoCode in streaming mode", () => {
+  it("should re-render code blocks when their content changes in place", async () => {
+    codeBlockRenderCount = 0;
+
+    const initial = "```js\nfirst\n```";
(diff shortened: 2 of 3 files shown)


dasheidt:
There is an adjacent half to this that content comparison alone will not reach: `CodeComponent`'s block branch never forwards its rest props to `CodeBlock` at all, while the inline branch four screens up does spread them. So an attribute a consumer puts on a `code` element via a rehype plugin reaches inline code and silently disappears on code blocks, whatever the comparator decides.

And once it is forwarded, `CodeBlockBody`'s comparator lists five props by hand and ignores the rest, so a forwarded attribute keeps the value it first rendered with even when `MemoCode` does re-render.

I opened #585 for those two, with a test that fails with only the forwarding half applied, so each is pinned separately. It is independent of this PR — happy to rebuild it on top if this lands first.

Worth noting that #584 (comparators generally) makes the in-place content case here fall out too, since a re-parsed block hands its `code` component new children. If that shape is the one the maintainers want, `sameCodeContent` may not be needed; if they prefer the targeted comparator, this PR stands on its own.

farnabaz:
Thanks for the PR. The issue was fixed in #584, but I picked your test in another PR (#593) and added you as a co-author.
Closing this.

PR #591, opened 2026-08-20
fix: allow remend incomplete-link sentinel through sanitize

### LINKED ISSUE #551: MarkdownA `data-incomplete` handling is unreachable: the built-in sanitize schema strips `streamdown:incomplete-link`, so default `linkMode: "protocol"` always degrades to the harden "[blocked]" replacement
### Summary

The graceful-render branch for incomplete streaming links in `MarkdownA` (`data-incomplete={isIncomplete}` keyed on `href === "streamdown:incomplete-link"`) is structurally dead code: the sentinel URL that remend injects can never reach the component, because Streamdown's own default rehype pipeline strips it first. The observable result is that the DEFAULT `remend.linkMode: "protocol"` flashes `{label} [blocked]` on every incomplete link mid-stream — the exact symptom of #156 — and the intended `data-incomplete` styling path never fires.

This is distinct from #156 / #165 / #301 (all closed): those report the flash or ask for an escape hatch (resolved by the opt-in `remend.linkMode: "text-only"`). This report is that the **default mode's own healing mechanism is internally unreachable**, so the default is structurally broken rather than merely opinionated.

### Mechanism (all refs on `main`, reproduces on 2.5.0)

1. **remend** (string level, streaming mode) completes a partial `[label](ur…` or `see [the docs` tail into `[label](streamdown:incomplete-link)` — `linkMode: "protocol"` default.
2. **rehype-sanitize** runs with the built-in schema (`packages/streamdown/index.tsx` ~L239-247), which extends `defaultSchema.protocols.href` with only `"tel"`. `defaultSchema.protocols.href` is `['http','https','irc','ircs','mailto','xmpp']` — **`streamdown:` is not in the list, so the sentinel `href` is deleted**.
3. **rehype-harden** (pipeline order `raw → sanitize → harden`, `index.tsx` ~L251-257) then sees an href-less `<a>`; `transformUrl(undefined, …)` returns `null` → the default `indicator` block policy **replaces the anchor** with `<span title="Blocked URL: undefined" …>{label} [blocked]</span>`. (Harden's own `allowedProtocols: ["*"]` would have allowed the sentinel — the block trigger is the sanitize-deleted href, not the protocol.)
4. By render time the node is a `<span>`, so `components.a` / `MarkdownA` is never invoked for it — the `isIncomplete` branch (`packages/streamdown/lib/components.tsx` ~L273, `data-incomplete` at ~L322/L344) cannot execute for streamed-incomplete links.

There is also no user-side escape at the schema layer: `allowedTags` merges `tagNames`/`attributes` only (no `protocols` extension), and replacing `rehypePlugins` wholesale disables the `allowedTags` mechanism.

### Suggested fix (either restores the designed behavior)

- Add the sentinel protocol to the built-in sanitize schema: `href: [...(defaultSchema.protocols?.href ?? []), "tel", "streamdown"]` — one line; harden already allowlists `*` protocols, and `MarkdownA` renders the sentinel case with `data-incomplete` + no navigable href, so nothing user-navigable leaks; **or**
- flip the default `linkMode` to `"text-only"` so the default never emits a sentinel it immediately destroys.

### Repro

```tsx
import { Streamdown } from "streamdown"; // 2.5.0
renderToStaticMarkup(<Streamdown mode="streaming">{"see [the docs"}</Streamdown>);
// → …see <span title="Blocked URL: undefined" class="text-gray-500">the docs [blocked]</span>
// expected (per MarkdownA's isIncomplete branch): an <a data-incomplete="true"> styled as pending
```

Workaround we ship in production: `remend={{ linkMode: "text-only" }}` — works as documented; this issue is only about the default mode's unreachable branch.

BetterAndBetterII:
Default sanitize dropped the remend streamdown:incomplete-link href, so streamed incomplete links always became harden [blocked]. Allow the streamdown protocol so MarkdownA data-incomplete can run. Fixes #551.

### DIFF
--- a/.changeset/fix-incomplete-link-sanitize-protocol.md
+++ b/.changeset/fix-incomplete-link-sanitize-protocol.md
@@ -1,0 +1,5 @@
+---
+"streamdown": patch
+---
+
+Allow remend's `streamdown:` incomplete-link sentinel through the default sanitize schema so streamed incomplete links render as pending instead of harden `[blocked]`.
--- a/packages/streamdown/__tests__/incomplete-link-sanitize.test.tsx
+++ b/packages/streamdown/__tests__/incomplete-link-sanitize.test.tsx
@@ -1,0 +1,53 @@
+import { render } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import { defaultRehypePlugins, Streamdown } from "../index";
+import { Markdown } from "../lib/markdown";
+
+const rehypePlugins = Object.values(defaultRehypePlugins);
+
+describe("Incomplete streaming links (#551)", () => {
+  it("should keep the remend sentinel href through the default sanitize schema", () => {
+    const { container } = render(
+      <Markdown
+        children="[the docs](streamdown:incomplete-link)"
+        rehypePlugins={rehypePlugins}
+      />
+    );
+    const link = container.querySelector("a");
+    expect(link).toBeTruthy();
+    expect(link?.getAttribute("href")).toBe("streamdown:incomplete-link");
+    expect(link?.textContent).toBe("the docs");
+    expect(container.textContent).not.toContain("[blocked]");
+  });
+
+  it("should render incomplete markdown links via MarkdownA instead of harden [blocked]", () => {
+    const { container } = render(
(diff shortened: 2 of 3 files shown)


farnabaz (approved):
Thanks

PR #603, opened 2026-08-31
chore(deps): update marked to v18

### LINKED ISSUE #566: Update `marked` to v18
Release notes: https://github.com/markedjs/marked/releases

Code:
https://github.com/vercel/streamdown/blob/e5deed330aa4231751a106445d93d62e4716a22f/packages/streamdown/package.json#L67

sleitor:
Closes #566

Bumps `marked` from `^17.0.1` to `^18.0.11` (latest) in `packages/streamdown/package.json`.

## Breaking change in marked v18

marked v18 no longer absorbs a block token's trailing blank line(s) into its own `raw`. That whitespace now surfaces as a separate `space` token immediately following the block (observed on `html`, `heading`, and `table` block tokens).

Example, for `<ai-thinking>\n\n**bold**\n\n</ai-thinking>\n\n# Outside`:

- v17: `html('<ai-thinking>\n\n')`, `paragraph('**bold**')`, `space('\n\n')`, `html('</ai-thinking>\n\n')`, `heading('# Outside')`
- v18: `html('<ai-thinking>')`, `space('\n\n')`, `paragraph('**bold**')`, `space('\n\n')`, `html('</ai-thinking>')`, `space('\n\n')`, `heading('# Outside')`

The `token.raw` lossless concatenation invariant still holds (joining all `raw` values still reconstructs the source exactly) — the difference is purely in how the whitespace is bucketed into tokens.

## Fix

`parseMarkdownIntoBlocks` (`packages/streamdown/lib/parse-blocks.tsx`) tracks an `htmlStack` to merge tokens between a custom tag's opening and closing HTML block tokens into one block. Under v17, the closing tag's trailing blank line was absorbed into that same `html` token, so it was included in the merge before the stack was popped. Under v18, that blank line is a separate `space` token emitted *after* the closing tag pops the stack, so it fell through and got pushed as its own standalone block — turning what should be 2 blocks into 3.

Since a bare `space` token is never meaningful content on its own, the fix folds any `space` token into the previous block (when one exists) instead of pushing it as a new block:

```ts
if (token.type === "space" && mergedBlocksLen > 0) {
  mergedBlocks[mergedBlocksLen - 1] += currentBlock;
  continue;
}
```

This restores block boundaries/counts identical to v17 output, for both the htmlStack-merge path and the general case. The existing math-block merge logic (`$$` odd-count tracking) and `previousTokenWasCode` tracking are untouched and placed after this check.

A regression test was added in `__tests__/custom-tag-markdown.test.tsx` asserting the v18 tokenization shape doesn't create an extra block and that `blocks.join("") === input` (lossless invariant).

## TypeScript

marked v18's internal TS toolchain bumped to v6, but this is not a consumer-facing constraint — TypeScript 5.9.3 (what this repo pins) typechecks cleanly against marked@18's `.d.ts` with `skipLibCheck: false`. No TS upgrade needed.

## Tests

`pnpm test` in `packages/streamdown`: all 1145 tests pass (1144 pre-existing + 1 new regression test), 84 test files.

## Changeset

Added `.changeset/update-marked-v18.md` as a `patch` for the `streamdown` package (dependency bump + internal parser fix, no public API change).

### DIFF
--- a/.changeset/update-marked-v18.md
+++ b/.changeset/update-marked-v18.md
@@ -1,0 +1,7 @@
+---
+"streamdown": patch
+---
+
+Update `marked` from `^17.0.1` to `^18.0.11`.
+
+marked v18 no longer folds a block token's trailing blank line(s) into its own `raw`; that whitespace now surfaces as a separate `space` token immediately following the block (e.g. after `html`, `heading`, and `table` tokens). This changed the token stream shape that `parseMarkdownIntoBlocks` consumes, causing a dangling `space` token after a closed custom-tag HTML block to be emitted as its own standalone block. `parseMarkdownIntoBlocks` now folds a `space` token into the preceding block instead of pushing it as a new one, restoring v17-identical block boundaries and counts. The lossless `token.raw` concatenation invariant is preserved.
--- a/packages/streamdown/__tests__/custom-tag-markdown.test.tsx
+++ b/packages/streamdown/__tests__/custom-tag-markdown.test.tsx
@@ -149,4 +149,27 @@ Content for snippet 2
     expect(blocks[0]).toContain("</ai-thinking>");
     expect(blocks[1]).toContain("# Outside");
   });
+
+  it("does not create a standalone block from a trailing blank-line space token (marked v18)", () => {
+    // marked v18 no longer folds a block token's trailing blank line(s) into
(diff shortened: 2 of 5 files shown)


farnabaz (approved):
Thanks 👍

PR #611, opened 2026-09-05
feat: support disableAutolinkProtocols to opt out of GFM autolinks per protocol

### LINKED ISSUE #607: Feature Request: Support disabling autolinks for specified URL protocols
### Problem Statement

Need a way to turn off autolink for certain protocols like mailto. Maybe add a prop like disableAutolinkProtocols={['mailto']}.
Thanks.

### Proposed Solution

Add a new prop (e.g. disableAutolinkProtocols) to Streamdown that accepts an array of URL protocols to disable autolinking for.
Internally, customize or extend the remark plugin (remark-gfm) to skip autolinking for the specified protocols (e.g. mailto).
This gives users fine-grained control over which protocols get autolinked.

### Alternatives Considered

_No response_

### Use Case

```tsx

```

### Priority

_No response_

### Contribution

### Additional Context

_No response_

sleitor:
Closes #607.

## What

Adds an optional `disableAutolinkProtocols` prop to `<Streamdown>` that disables GFM autolink-literal linking for specific URL protocols (e.g. `mailto`), as requested in #607.

```tsx
<Streamdown disableAutolinkProtocols={["mailto"]}>
  {"Contact us at hello@example.com"}
</Streamdown>
```

Before:

```html
<p>Contact us at <a href="mailto:hello@example.com">hello@example.com</a></p>
```

After (with `disableAutolinkProtocols={["mailto"]}`):

```html
<p>Contact us at hello@example.com</p>
```

- Protocol names are case-insensitive and accept an optional trailing colon (`"mailto"` and `"mailto:"` are equivalent).
- Only literal autolinks created by `remark-gfm`'s autolink-literal extension are affected (bare emails, bare `http(s)`/`www` URLs). Explicit markdown links (`[text](mailto:...)`) are left as links, since they're an intentional link, not an autolink.
- **Default behavior is completely unchanged** when the prop is omitted — no extra plugin is added to the remark pipeline in that case.

## How

`mdast-util-gfm-autolink-literal` doesn't tag the link nodes it creates, so a small remark plugin (`packages/streamdown/lib/remark/disable-autolink-protocols.ts`) identifies them structurally: a link with exactly one `text` child whose value reconstructs the node's `url` (accounting for the `mailto:` prefix GFM adds to bare emails). Matching nodes are unwrapped back to their text children. The plugin runs immediately after `remark-gfm` in the pipeline (before the CJK/math plugin slots), matching the existing "before/after gfm" plugin ordering convention.

It's exposed via the standard unified `[plugin, options]` tuple form rather than a plugin-factory closure — Streamdown's internal `Markdown` processor cache keys processors by plugin name + `JSON.stringify(options)`, and a factory returning a fresh anonymous closure per call would always serialize to the same cache key, causing different `disableAutolinkProtocols` configurations to silently reuse a stale cached processor (caught this via a failing test while developing).

## Testing

Added `packages/streamdown/__tests__/disable-autolink-protocols.test.tsx` covering:
- bare email not linked when `mailto` is disabled
- default behavior (prop omitted) unchanged
- explicit `[text](mailto:...)` links stay linked
- http/https autolinks still linked when only `mailto` is disabled
- case-insensitivity and the `"mailto:"` form
- disabling `https` leaves `mailto` autolinks untouched
- end-to-end wiring through the `<Streamdown disableAutolinkProtocols>` prop itself

`pnpm -F streamdown exec vitest run` passes except for 5 pre-existing failures on `main` unrelated to this change (image-placeholder / incomplete-image-sanitize tests, verified failing identically on a clean `upstream/main` checkout before my changes).

Also updated `apps/website/content/docs/configuration.mdx` with the new prop, and added a changeset.

### DIFF
--- a/.changeset/disable-autolink-protocols.md
+++ b/.changeset/disable-autolink-protocols.md
@@ -1,0 +1,15 @@
+---
+"streamdown": minor
+---
+
+Add a `disableAutolinkProtocols` prop to `<Streamdown>` for disabling GFM autolinking of specific URL protocols (e.g. `mailto`).
+
+```tsx
+<Streamdown disableAutolinkProtocols={["mailto"]}>
+  {"Contact us at hello@example.com"}
+</Streamdown>
+```
+
+Bare emails and bare URLs whose protocol matches the list (case-insensitive, `"mailto"` and `"mailto:"` are equivalent) are unwrapped back to plain text. Explicit markdown links (`[text](mailto:...)`) are left as links. When the prop is omitted, autolinking behavior is completely unchanged.
+
+Closes #607.
--- a/apps/website/content/docs/configuration.mdx
+++ b/apps/website/content/docs/configuration.mdx
@@ -129,6 +129,11 @@ With `dir="auto"`:
       type: "Pluggable[]",
       default: "Object.values(defaultRemarkPlugins)",
     },
+    disableAutolinkProtocols: {
+      description:
+        "Disable GFM autolinking for specific URL protocols (e.g. ['mailto'] stops bare email addresses from becoming links). Accepts protocol names with or without a trailing colon, case-insensitive. Only affects literal autolinks created by remark-gfm; explicit [text](url) links are unaffected.",
+      type: "string[]",
+    },
   }}
 />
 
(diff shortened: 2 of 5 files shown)


farnabaz (approved):
Thanks

Issue #616, opened 2026-09-10
remend `htmlTags` handler has no math guard: `\sum_{j<k}` deletes the rest of the message


WHQ25:
### Bug Description

`remend`'s `htmlTags` handler deletes everything from an unclosed `<tag` to the end of the string. Two things make it fire far outside its intended scope:

1. **It has no math guard.** The handler checks `isWithinCodeBlock` but never `isWithinMathBlock` — even though `remend` implements and exports the latter. A perfectly ordinary LaTeX expression like `\sum_{j<k}` therefore triggers it.
2. **It fires on complete text.** The regex is `/<[a-zA-Z/][^>]*$/`, i.e. "a `<` with no `>` anywhere after it". That heuristic is meaningful for the tail of a streaming chunk, but `remend(text)` has no way to know whether the stream has ended, so it applies the same cut to text that is already final.

The two combine into a bad failure mode for chat UIs: an assistant message that mentions `\sum_{j<k}` loses **its entire remainder**. And because the `katex` handler (priority 70) runs after `htmlTags` (priority 10), the now-orphaned `$$` gets auto-closed, so KaTeX receives truncated input and renders a `katex-error` in `--color-muted-foreground`. The user sees the answer stop mid-formula in grey text, with no error anywhere. It took a while to trace that back to markdown repair rather than to the model or to KaTeX.

This is not math-specific — any prose where `<` is followed by a letter and no `>` appears later is affected, e.g. `the loop runs while i<n`.

The relevant code in `packages/remend`:

```js
const HTML_TAG_RE = /<[a-zA-Z/][^>]*$/

const handleHtmlTags = (text) => {
  const match = text.match(HTML_TAG_RE)
  return !match || match.index === undefined || isWithinCodeBlock(text, match.index)
    ? text
    : text.substring(0, match.index).trimEnd()
}
```

This looks like the same class of gap as #1 (`handleIncompleteSingleUnderscoreItalic` not math-aware) and #522 (emphasis handlers not recognising `\(...\)` / `\[...\]`). Those were fixed for the emphasis handlers; `htmlTags` was never covered.

### Steps to Reproduce

```js
import remend from 'remend'

remend('前文\n\n$$\nI = \\sum_{j<k} p_j\n$$\n\nTAIL')
// → '前文\n\n$$\nI = \\sum_{j\n$$'          TAIL is gone, math is truncated

remend('inline $$A_{j<k}$$ more\n\nTAIL')
// → 'inline $$A_{j$$'

[...]

PR #617, opened 2026-09-10
fix(remend): guard incomplete HTML tag handler against math blocks

torsello:
## Description

`handleIncompleteHtmlTag` guarded against code blocks but never against math, so an ordinary
comparison inside a math expression matched the incomplete-tag pattern and deleted everything
from the `<` to the end of the string.

```js
remend('Intro\n\n$$\nI = \\sum_{j<k} p_j\n$$\n\nTAIL')
// before → 'Intro\n\n$$\nI = \\sum_{j\n$$'   ← TAIL gone, math truncated
// after  → unchanged
```

Because the `katex` handler (priority 70) runs after `htmlTags` (priority 10), the now-orphaned
`$$` was auto-closed, so KaTeX received unterminated input and rendered a `ParseError` in
`--color-muted-foreground`. The net effect for the reader is a message that silently ends
mid-formula in grey text, with no error surfaced anywhere.

`remend` already implements and exports `isWithinMathBlock`, and the emphasis handlers already
use it — `htmlTags` was simply never covered.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Related Issues

Related to #616

<sub>Deliberately not `Fixes` — this covers the math half of that issue. See **Additional Notes**.</sub>

## Changes Made

- Added an `isWithinMathBlock` guard to `handleIncompleteHtmlTag`, gated behind the same
  `hasMathDelimiters` fast path the emphasis handlers use, so the `O(n)` scan is skipped
  entirely for text with no math delimiters.
- **Changed the handler to walk forward through candidates instead of bailing out on the
  first guarded one.** This is the non-obvious half. `incompleteHtmlTagPattern` is
  leftmost-matching and its `[^>]*$` tail means the match always runs to the end of the
  string — so every later `<` that starts a tag name is an equally valid candidate. Simply
  returning `text` when the first match falls inside math would trade one bug for another:

  ```js
  remend('$x<y$ then <div')
  // guard-only → '$x<y$ then <div'   ← the genuine incomplete tag now survives. wrong.
  // this PR    → '$x<y$ then'        ← correct
  ```

  The same walk-forward now applies to code blocks, which fixes a smaller unreported case
  on its own: a `<` inside a fenced block no longer masks a real incomplete tag after it.
- Added 4 test cases to `packages/remend/__tests__/html-tags.test.ts`.
- Added a patch changeset for `remend`.

## Testing

- [x] All existing tests pass

New cases cover block `$$`, inline `$`, both LaTeX delimiter forms (`\(…\)` and `\[…\]`), and
the walk-forward behaviour. I confirmed all four fail on `main` before the fix and pass after —
the walk-forward case in particular fails against a guard-only implementation too.

Manual verification against the exact reproductions from the issue:

| Input | Before | After |
| --- | --- | --- |
| `Intro\n\n$$\nI = \sum_{j<k} p_j\n$$\n\nTAIL` | truncated at `\sum_{j` | unchanged |
| `inline $$A_{j<k}$$ more\n\nTAIL` | `inline $$A_{j$$` | unchanged |
| `Hello <div` | `Hello` | `Hello` (unchanged behaviour) |

### Test Coverage

`remend`: 493 tests across 26 files, all passing.
Full monorepo suite: 1172 tests across 85 files, 6/6 turbo tasks green.
`biome check` reports no findings on both touched files.

<sub>Note: `tsc --noEmit` reports two `TS18048` errors in `packages/remend/src/strikethrough-handler.ts`.
I verified these are present on `main` and unrelated to this change.</sub>

## Screenshots/Demos

Not applicable — no visual surface changes. The before/after is shown as code above.

## Checklist

<sub>Documentation left unchecked: this is a behavioural fix with no public API change, so there is
nothing in the docs to update. Happy to add a note to the termination docs if you'd like one.</sub>

## Changeset

`.changeset/html-tag-math-guard.md` — `patch` bump for `remend`.

## Additional Notes

**What this PR intentionally leaves out.** #616 raises a second, broader problem: the handler
also fires on text that is already complete, so ordinary prose like `the loop runs while i<n`
still loses its remainder. That one is not a missing guard — it needs an explicit signal,
because `remend(text)` currently cannot distinguish a streaming tail from settled text, and
every tail-oriented heuristic keeps firing after the stream ends.

The issue suggests an option or a `mode` argument mirroring `Streamdown`'s `mode="static"`.
That is a public API decision affecting more than this handler, so I did not want to make it
unilaterally in a first contribution. Happy to open a follow-up if you want to shape the API —
or to fold it into this PR if you already know which direction you'd prefer.

### DIFF
--- a/.changeset/html-tag-math-guard.md
+++ b/.changeset/html-tag-math-guard.md
@@ -1,0 +1,16 @@
+---
+"remend": patch
+---
+
+Stop the incomplete-HTML-tag handler from truncating math expressions.
+
+`handleIncompleteHtmlTag` guarded against code blocks but not math, so an ordinary
+comparison inside math — `$$ I = \sum_{j<k} p_j $$` — matched the incomplete-tag
+pattern and deleted everything from the `<` to the end of the string. The trailing
+`$$` was then auto-closed by the katex handler, so KaTeX rendered a parse error and
+the rest of the message never reached the DOM.
+
+The handler now skips candidates inside math blocks (`$`, `$$`, `\(`, `\[`), matching
+the guard the emphasis handlers already had. Because the pattern is leftmost-matching,
+it also walks forward to later candidates instead of bailing out, so a genuine
+incomplete tag after a math expression is still stripped.
--- a/packages/remend/__tests__/html-tags.test.ts
+++ b/packages/remend/__tests__/html-tags.test.ts
@@ -74,6 +74,29 @@ describe("incomplete HTML tag stripping", () => {
     );
   });
 
+  it("should not strip < inside block math", () => {
+    const text = "Intro\n\n$$\nI = \\sum_{j<k} p_j\n$$\n\nTAIL";
+    expect(remend(text)).toBe(text);
+  });
+
+  it("should not strip < inside inline dollar math", () => {
+    const text = "inline $A_{j<k}$ more";
+    expect(remend(text)).toBe(text);
+  });
+
+  it("should not strip < inside LaTeX delimiters", () => {
+    const inline = "value \\(a<b\\) end";
+    expect(remend(inline)).toBe(inline);
+
+    const block = "value \\[a<b\\] end";
+    expect(remend(block)).toBe(block);
+  });
+
+  it("should still strip incomplete tags outside math blocks", () => {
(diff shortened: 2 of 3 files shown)


farnabaz (approved):
LGTM 👍 
Thanks

PR #618, opened 2026-09-12
fix: preserve literal HTML in Markdown table exports

lllleolin-max:
## Description

Copying or downloading a table as Markdown can change its cell text when the export is rendered again. For example, `Array<string>` becomes `Array`, and the literal text `&copy;` becomes `©`.

Escape `&`, `<`, and `>` in the shared Markdown cell serializer so those values survive export. The existing single-pass serializer still emits `<br>` for actual newlines.

## Type of Change

## Changes Made

- Preserve literal HTML and character entities in Markdown table exports, including headers.
- Add a DOM → Markdown → DOM regression test using the existing `marked` dependency.
- Add a patch changeset for `streamdown`.

## Testing

- [x] All existing tests in the changed package pass: `pnpm --filter streamdown test` — 85 files, 1,173 tests.
- [x] Confirmed the new regression fails on the original serializer and passes with the fix (47 table utility tests).
- [x] `pnpm -r --filter 'streamdown...' build` — core and dependency packages build, including declarations.
- [x] `pnpm check` — 291 files checked.
- [ ] Manually tested the changes

An extra `pnpm --filter streamdown exec tsc --noEmit` check reports 211 existing diagnostics on both the original and patched source; diagnostics match after normalizing source positions. This check is not passing on the baseline either.

## Checklist

- [ ] I have performed a self-review of my own code
- [x] New and existing unit tests in the changed package pass locally
- [x] I have created a changeset

## Changeset

## Additional Notes

Prepared with OpenAI Codex. Automated tests and Codex agent review were completed; no independent human review is claimed.


farnabaz (approved):
Thanks

PR #619, opened 2026-09-12
fix: preserve streaming animations across text and code

rogersnm:
## Description

Streaming updates can truncate active fades, completion can reveal pending text instantly, replay can skip opening animations, generated list whitespace can make new words appear already displayed, and code blocks can appear ahead of the text before them. Preserve animation history and use the shared animation timeline for text and code.

## Type of Change

- [x] Bug fix

## Changes Made

- Preserve each pending or active fade's original timing across updates, including list markers, checkboxes, images and rules.
- Let finite Streamdown CSS animations finish after streaming stops, then remove animation attributes/wrappers. Cleanup uses `Animation.finished`, handles cancellation, and is invalidated when streaming resumes or the final content changes.
- Discard removed blocks' history and reset the timeline when content is cleared, so replay animates its opening words.
- Exclude positionless parser-generated layout whitespace from both animation counting passes. Source whitespace still counts; loose-list and tight-to-loose transitions preserve offsets.
- Animate fenced code as one unit on the shared timeline. Its entry fade follows preceding words without inserting spans into syntax-highlighted content. A stable wrapper preserves the fade while code and highlighting update and while final animations drain. All blocks share one render-pass clock, so parsing time cannot let a later block overtake earlier text.

Animation defaults and the public API are unchanged. Code uses the configured duration and timeline; it does not get per-token syntax animations. Existing callbacks retain their streaming-state semantics.

## Testing

- Streamdown: **1,206 passing tests**, including plugin-level code scheduling and a React regression for the persistent code wrapper.
- `pnpm check` passes.
- The patched 2.6.0 distribution builds successfully and passes **111 consumer Markdown/message-bubble tests**.
- **14 consumer browser checks** pass across Chromium and WebKit: pending/active fades, completion/replay, list offsets, asynchronous plugin updates, and code entry ordering, growth and completion. The browser fixture uses real CSS animations and the actual installed component.
- Full-response playback also verified in both engines, with cold and cached maths loading.

The existing package/website build and six-package test validation from the earlier commits remains described in the linked history. The optional `apps/test` production build requires an AI Gateway credential; it was not used as a gate.

## Screenshots/Demos

[Original before/after animation recording](https://github.com/rogersnm/streamdown/blob/demo/animation-lifetime/comparison.mp4) · [Quarter-speed recording](https://github.com/rogersnm/streamdown/blob/demo/animation-lifetime/comparison-slow.mp4)

The original recording covers the update/completion/replay fixes; it predates the list-offset and code-entry additions. Its 50 ms updates, 250 ms fades and 10 ms stagger are comparison settings, not changed defaults.

List reproduction: render `1. First\n\n2. *But in ourselves, that we`, then append ` are underlings.`. Previously “are” could receive a zero-duration fade while adjacent words remained mid-animation.

Code reproduction: deliver a heading and fenced code in the same update with staggered animation enabled. Previously the code bypassed the timeline; now its wrapper receives one slot after the heading and keeps the same fade when the fence grows.

The consumer's separate fix for changing its React key when its maths plugin loads is application-side and is not included here.

## Checklist

- [x] Followed project code style.
- [x] Added and updated regression tests.
- [x] Updated the patch changeset.

### DIFF
--- a/packages/streamdown/lib/components.tsx
+++ b/packages/streamdown/lib/components.tsx
@@ -1112,12 +1112,13 @@ export const components: Options["components"] = {
   blockquote: MemoBlockquote,
   code: MemoCode,
   img: MemoImg,
-  pre: ({ children }) => {
-    if (isValidElement(children)) {
-      return cloneElement(children, { "data-block": "true" });
-    }
-    return children;
-  },
+  pre: ({ children, node: _node, ...props }) => (
+    <div {...props}>
+      {isValidElement(children)
+        ? cloneElement(children, { "data-block": "true" })
+        : children}
+    </div>
+  ),
   sup: MemoSup,
   sub: MemoSub,
   p: MemoParagraph,
--- a/.changeset/tidy-words-fade.md
+++ b/.changeset/tidy-words-fade.md
@@ -1,0 +1,9 @@
+---
+"streamdown": patch
+---
+
+Preserve pending and active word animations across streaming updates instead of setting their duration to zero before they finish. Apply the same timing to list markers, task checkboxes, images, and horizontal rules.
+
+Allow finite text animations to finish after streaming stops before removing their wrappers. Cleanup follows browser animation completion, including cancellation, rather than using a fixed timeout.
+
+Discard animation history for removed blocks and reset the timeline when content is cleared, so the first words of a replay animate again instead of appearing instantly.
--- a/.changeset/tidy-words-fade.md
+++ b/.changeset/tidy-words-fade.md
@@ -9,3 +9,5 @@ Allow finite text animations to finish after streaming stops before removing the
 Discard animation history for removed blocks and reset the timeline when content is cleared, so the first words of a replay animate again instead of appearing instantly.
 
 Exclude parser-generated layout whitespace from animation offsets so new words inside lists receive their fade and tight-to-loose list transitions preserve existing text history. Source whitespace remains counted.
+
+Animate fenced code as one unit on the shared timeline, preserving its fade while code and syntax highlighting update.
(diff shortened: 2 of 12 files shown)


packages/streamdown/lib/components.tsx
```diff
@@ -1112,12 +1112,13 @@ export const components: Options["components"] = {
...
-    }
-    return children;
-  },
+  pre: ({ children, node: _node, ...props }) => (
+    <div {...props}>
+      {isValidElement(children)
+        ? cloneElement(children, { "data-block": "true" })
+        : children}
+    </div>
+  ),
```
farnabaz, line 1121: Instead of having an extra wrapper everytime, lets add wrapper once it needed 
```ts
pre: ({ children, node: _node, ...props }) => {
  const wrapperProps = {
    ...props,
    className: [props.className, children.props.className].filter(Boolean).join(" ") || undefined,
    style: { ...props.style, ...children.props.style },
  }

  if (!isValidElement(children)) {
    return <div {...wrapperProps}>{children}</div>;
  }

  return cloneElement(children, { ...wrapperProps, "data-block": "true" });
},

```

rogersnm: Thanks, I looked at moving the props onto the child, but the wrapper is doing real work here. `CodeBlock` forwards extra props to the body rather than the container, and the body inside `Suspense` gets swapped from the plain fallback to the highlighted version when highlighting resolves. That new element restarts the fade, and the header and actions wouldn't fade at all. The Mermaid and custom-renderer paths don't forward props, so those would lose the fade entirely. Merging `className` and `style` also makes new objects every render, which defeats `MemoCode`'s memo check.

Wrapping only while animating doesn't work either: when the drain finishes and the animate plugin turns off, the wrapper would disappear and remount the code block, so highlighting runs again and Mermaid diagrams re-render. A wrapper that's always there keeps the tree the same. It has no box styles, so layout doesn't change. The test "keeps the animated code wrapper when a streamed fence grows" covers the stable-element part. Happy to add a short comment on `pre` explaining why the wrapper is there.

Issue #326, opened 2026-01-04
Autolink literals swallow CJK punctuation and following text


xxchan:
### Bug Description
Autolink literals in CJK text treat fullwidth punctuation as part of the URL when there is no whitespace. Example: `请访问 https://example.com。谢谢` becomes a link to `https://example.com。谢谢`, so the punctuation and following CJK text are swallowed by the link.

### Steps to Reproduce
```tsx
import { Streamdown } from "streamdown";

export default function App() {
  const markdown = "请访问 https://example.com。谢谢";
  return <Streamdown>{markdown}</Streamdown>;
}
```

### Expected Behavior
- Link text/href: `https://example.com`
- Trailing text `。谢谢` remains plain text

### Actual Behavior
- Link text/href includes the trailing CJK punctuation and text: `https://example.com。谢谢`

### Environment
- streamdown: 1.6.10
- remark-gfm: 4.0.1 (default)

### Additional Context
CJK emphasis is handled via remark-cjk-friendly, but autolink literal boundaries still follow GFM behavior. This feels related to trailing punctuation issues in autolinks (e.g. remarkjs/remark-gfm#12), but fullwidth CJK punctuation is not treated as a boundary.

PR #624, opened 2026-09-19
fix(cjk): re-linkify URLs after CJK punctuation splits

### LINKED ISSUE #623: CJK autolink boundary handling leaves subsequent URLs as plain text
### Bug Description

With @streamdown/cjk enabled, multiple bare URLs separated by fullwidth semicolons (；, U+FF1B) without whitespace render with only the first URL clickable. Subsequent URLs remain plain text. This appears to be an uncovered case in #327 (fixing #326), rather than an error in the boundary index calculation.

### Steps to Reproduce

Render the code sample below with mode="static", then with mode="streaming". In both cases only the first URL becomes a link. Adding a space after each semicolon, or writing three explicit Markdown links, makes all three links work.

### Expected Behavior

All three URLs should be separate clickable links, with the semicolons preserved as plain text.

### Actual Behavior

Only https://example.com/a is clickable. The rest is plain text. With a custom a renderer to inspect the HTML:

```html
<p>参考资料：<a href="https://example.com/a">https://example.com/a</a>；https://example.com/b；https://example.com/c</p>
```

### Code Sample

```tsx
import { cjk } from '@streamdown/cjk';

const markdown =
  '参考资料：https://example.com/a；https://example.com/b；https://example.com/c';

  return (
    <Streamdown mode="static" plugins={{ cjk }}>
      {markdown}
    </Streamdown>
  );
}
```

### Streamdown Version

2.5.0 and 2.6.0; @streamdown/cjk 1.0.3

### React Version

19.2.4

### Node.js Version

24.13.1

### Browser(s)

_No response_

### Operating System

None

### Additional Context

The following diagram illustrates the observed parsing result; it is not a browser screenshot.

```mermaid
flowchart TD
  input["Input: URL A；URL B；URL C"] --> gfm["GFM: one link containing A；B；C"]
  gfm --> boundary["CJK boundary: split at the first ；"]
  boundary --> actual["Actual: link A + plain text ；B；C"]
  input -. "Expected rendering" .-> expected["link A + ； + link B + ； + link C"]
  style boundary fill:#fff3cd,stroke:#b58105,color:#222
  style actual fill:#ffe8e8,stroke:#c74343,color:#222
  style expected fill:#e7f6ec,stroke:#348450,color:#222
```

In #327, remarkCjkAutolinkBoundary finds the first CJK punctuation boundary and replaces the original link with:

```ts
const trimmedLink = buildAutolink(trimmedUrl, node);
const trailingText = buildTrailingText(trailing);
parent.children.splice(index, 1, trimmedLink, trailingText);
```

For `URL。谢谢`, this correctly preserves trailing prose as text. For `URL A；URL B；URL C`, the trailing part contains two more URLs. It becomes a text node and is not re-linkified by this link-node visitor.

Without the boundary handling, GFM also does not produce three correct links: it treats the sequence as one long URL. This report is about preserving subsequent URLs while applying the CJK boundary fix, not claiming the pre-#327 output was correct.

Verification:
- Reproduced with Streamdown 2.5.0 and @streamdown/cjk 1.0.3.
- Also tested the original three-URL message after upgrading to Streamdown 2.6.0 with @streamdown/cjk 1.0.3, without our local workaround: the same two static/streaming regression tests fail.
- Adding a space after each semicolon makes all three links render correctly.
- Writing three explicit Markdown links also works.
- Tests used React server rendering with complete content in both modes; incremental token delivery was not tested.

Could the CJK autolink handling preserve or re-detect subsequent bare URLs after splitting at punctuation? A regression test for multiple URLs separated by fullwidth semicolons would cover this case.

vyrnsynx:
## Description

Fixes GFM autolink literals that contain multiple bare URLs separated by CJK punctuation (e.g. fullwidth semicolons). After the existing boundary split from #327, trailing segments that contain further URLs are walked and re-linkified instead of left as plain text.

## Type of Change

## Related Issues

Fixes #623
Related to #327 / #326

## Changes Made

- Expand `remarkCjkAutolinkBoundary` to walk the full autolink string after each CJK split
- Preserve intervening CJK punctuation as text nodes
- Add regression tests for fullwidth-semicolon URL lists and URL-after-ideographic-full-stop
- Add changeset for `@streamdown/cjk` (patch)

## Testing

- [x] All existing tests pass
- [ ] Manually tested the changes

### Test Coverage

```
pnpm --filter @streamdown/cjk test
```

27 tests passed (including 2 new regressions for #623).

## Checklist

## Changeset


farnabaz (approved):
Thanks

